Issue #46, opened 2024-02-26
Unit tests do not work on Intel GPUs


mirenradia:
## Summary

The unit tests do not work on Intel GPUs. I think there seems to be some incompatibility between Catch2 and the Intel DPC++ compiler's SYCL implementation.

## Steps to reproduce

Here are some steps to reproduce on [Dawn](https://www.hpc.cam.ac.uk/d-w-n) at its state on 2024-02-26 (likely to change soon): 

1. SSH to Dawn
2. Clone AMReX:
   ```bash
   git clone https://github.com/AMReX-Codes/amrex.git
   ```
3. Clone this repository:
   ```bash
   git clone https://github.com/GRTLCollaboration/GRTeclyn.git
   ```
4. Start an interactive job (it is not currently possible to build on the login nodes).
5. Load the required Intel modules:
   ```bash
   module load intel-oneapi-compilers/2024.0.0/gcc/znjudqsi intel-oneapi-mkl/2024.0.0/oneapi/4n7ruz44
   ```
6. Change into the Tests directory:
   ```
   cd GRTeclyn/Tests
   ```
7. Build with `USE_SYCL=TRUE`:
   ```bash
   make -j <num jobs>  USE_SYCL=TRUE
   ```
8. Run the tests
   ```bash
   ./Tests3d.sycl.ex
   ```

## Observed outcome

The tests abort with the following error:
```
amrex::Abort::0::ParallelFor: Native API failed. Native API returns: -30 (PI_ERROR_INVALID_VALUE) -30 (PI_ERROR_INVALID_VALUE)!!!!! !!!
SIGABRT
See Backtrace.0 file for details
```

## Expected outcome

The tests should work without issues.

## Additional information

I think there are several compounding issues here:

1. The Intel DPC++ compiler has trouble linking unnamed device kernels in Catch2 test cases. See intel/llvm#10659. 

   Even if I remove all but one of the test cases with `amrex::ParallelFor`s such as the "CCZ4 RHS" test case by modifying `test_dirs` in [Tests/GNUMakefile](https://github.com/GRTLCollaboration/GRTeclyn/blob/41e41c772897516fa35a4f53865df83fa72b82d7/Tests/GNUmakefile) to just
   ```makefile
   test_dirs = $(GRTECLYN_TESTS_HOME)/Common \
               $(GRTECLYN_TESTS_HOME)/CCZ4RHSTest
   ```
[...]

PR #49, opened 2024-03-12
Tests: Switch from Catch2 to doctest

### LINKED ISSUE #47: Switch unit test framework from Catch2 to doctest
We should investigate switching from Catch2 to [doctest](https://github.com/doctest/doctest) as preliminary tests suggest it may not suffer from some of the problems Catch2 has mentioned in #46. 
In particular, it seems to be fine with my minimal reproducer for intel/llvm#10659. See the [doctest branch of my Catch2-SYCL repo](https://github.com/mirenradia/Catch2-SYCL/tree/doctest).

### LINKED ISSUE #48: Unit tests do not work on AMD GPUs
## Summary

Like #46 for Intel GPUs, there seems to be problems with our unit tests on AMD GPUs at least on the version of ROCm (5.1.0) and specific GPU that I tried (I think MI210).

## Steps to reproduce

Here are some steps to reproduce on the [AMD GPU nodes on COSMA](https://cosma.readthedocs.io/en/latest/gpu.html).

1. SSH to COSMA8
2. Clone AMReX:
   ```bash
   git clone https://github.com/AMReX-Codes/amrex.git
   ```
3. Clone this repo:
   ```bash
   git clone https://github.com/GRTLCollaboration/GRTeclyn.git
   ```
5. SSH to `ga005` (which is running ROCm v5.1.0 at the time of writing)
6. ROCm 5.1.0 is already in the system path so there is no need to load any modules for it but it is necessary to load a newer version of GCC and tell `hipcc` to use it. First load the module
   ```bash
   module load gnu_comp/11.1.0
   ```
7. Add the following lines to `~/amrex/Tools/GNUMake/Make.local` (create this file if it doesn't exist):
   ```make
   ifeq ($(USE_HIP),TRUE)
     GCC_PATH := $(shell realpath -m $(shell which gcc)/../..)
     GCC_VERSION := $(notdir $(GCC_PATH))
     CXXFLAGS += --gcc-toolchain=$(GCC_PATH)
     SYSTEM_INCLUDE_LOCATIONS += $(GCC_PATH)/include/c++/$(GCC_VERSION)
     AMREX_AMD_ARCH = gfx90a
   endif
   ```
8. Change into the Tests directory:
   ```bash
   cd ~/GRTeclyn/Tests
   ```
9. Build with `USE_HIP=TRUE`:
   ```bash
   make -j 128 USE_HIP=TRUE
   ```
10. Run the tests
    ```bash
    make run USE_HIP=TRUE
    ```

## Observed outcome

The tests abort with the following error:
```
:0:rocdevice.cpp            :2614: 1823894239190 us: 124592: [tid:0x2b2d6af70700] Device::callbackQueue aborting with error : HSA_STATUS_ERROR_MEMORY_FAULT: Agent attempted to access an inaccessible address. code: 0x2b
SIGABRT
```

## Additional information

Passing `-DCATCH_CONFIG_NO_COUNTER` to the preprocessor which changes the way Catch2 internally names test cases uniquely from using the `__COUNTER__` predefined macro to the `__LINE__` one allows the tests to work (although the CCZ4 RHS test currently fails as the tolerances are currently too small). 

Since the test cases are currently in different translation units (i.e. different `.cpp` files) which means `__COUNTER__` ends up being `0` for each of them, the "unique" names are not unique and the ROCm linker has difficulty figuring out which device code to link to each kernel in test cases. Since `TEST_CASE` is currently on a different line for each test (although we can't guarantee that going forward), `__LINE__` ends up being different and the test cases are uniquely named.

mirenradia:
This PR switches the unit testing framework from Catch2 to [doctest](https://github.com/doctest/doctest) which is a single header library and faster to compile than Catch2 (though a bit less feature rich). There is also an [MPI extension](https://github.com/doctest/doctest/blob/master/doc/markdown/extensions.md#distributed-tests-with-mpi) which we might want to investigate using in the future.

The [README](https://github.com/GRTLCollaboration/GRTeclyn/blob/enhancement/doctest/Tests/README.md) has been updated and provides documentation on how this works. Note that we no longer have to deal with git submodules and the doctest header is included directly in this repo.

This resolves #47 and fixes #48. Unfortunately #46 is not fixed by this (except when built with `DEBUG=TRUE`) so we'll leave that open for now.


julianakwan (dismissed):
Works great! I both built it and added a (dummy) example. I only have one minor suggestion: could you please add an extra line to the instructions so that users know they also have to `#include "doctestCLIArgs.hpp"` if they want to use `doctest::cli_args.argc()` and `doctest::cli_args.argv()`? Thanks!

julianakwan (changes requested):
I mistakenly selected 'approve' when it should have been 'request changes'. See above comment for changes.

julianakwan (approved):

PR #54, opened 2024-03-28
Fix null `m_gr_amr_ptr` on restart (#52)

### LINKED ISSUE #52: `GRAMRLevel::m_gr_amr_ptr` is null on restart
This leads to segfaults when this pointer is dereferenced e.g. in `GRAMRLevel::post_restart`.

mirenradia:
This fixes #52 by adding a getter for the `GRAMR` pointer stored by each `GRAMRLevel` object which sets the value if it's null. The pointer has now been made `private` so that derived classes use the getter rather than using the pointer directly.

I have also added a step to the regression test GH action workflow to restart from the last checkpoint and evolve for a few more steps to prevent this issue (or a similar one) from appearing again.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -67,3 +67,9 @@ jobs:
         --rel_tol 1e-10 \
         ${GITHUB_WORKSPACE}/GRTeclyn/.github/workflows/data/plt00008_compare \
         ${BINARYBH_EXAMPLE_DIR}/plt00008
+
+    - name: Restart from last checkpoint and evolve for a few more steps
+      run: >
+        mpiexec -n 2 ./main3d.gnu.MPI.ex ./params_test.txt
+        amr.restart=chk00008 max_steps=12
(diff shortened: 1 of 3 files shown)


julianakwan:
Hi Miren - the changes made to initialize m_gramr_pt look good to me but the checkpoint restarts are failing the  regression test. Do you want to split the pull request into two separate ones so that the pointer part can be merged while you work on the checkpoint restarts?

mirenradia:
> Hi Miren - the changes made to initialize m_gramr_pt look good to me but the checkpoint restarts are failing the regression test. Do you want to split the pull request into two separate ones so that the pointer part can be merged while you work on the checkpoint restarts?

I think the regression tests passed on the last commit. The point in the first commit was to show that they would currently fail but then the change in the final 2 commits fixes it.

julianakwan:
Thanks - I thought there 3 independent commits and only the third one passed.

julianakwan (approved):

PR #57, opened 2024-04-18
CI: Fix lint action and fail if warnings are present.

### LINKED ISSUE #12: Consider setting up clang-tidy

### LINKED ISSUE #56: CI: Lint/clang-tidy action is failing due to unknown `latest` version
Presumably, [cpp-linter/cpp-linter-action](https://github.com/cpp-linter/cpp-linter-action) removed the `latest` tag from their repo. We should update the workflow to use a numbered version.

mirenradia:
This PR does the following:
* Fixes #56 by using `v2` of the cpp-linter-action.
* Updates the lint workflow to fail if there are warnings.
* Updates the lint workflow to run on the `pull_request` trigger.
* Removes the style workflow and uses the clang-format checking provided by cpp-linter-action in the lint workflow.
* Updates the lint workflow to post a comment on PRs with the warnings outputted by clang-tidy and the incorrectly formatted files as determined by clang-format (an example can probably be found below).
* Updates the LLVM version used (for clang-tidy and clang-format) to 17 (the pre-commit config has also been updated).
* Updates the lint workflow to use a compilation database created by [compiledb](https://github.com/nickdiego/compiledb) rather than parsing the output of e.g. `make print-CXXFLAGS` and adding these as extra flags. Note that this means that files which are not in the compilation database (i.e. are not currently in a test or example) will not have any flags associated with them and there will be warnings.

Note that our `.impl.hpp` template implementation files will always fail linting as they cannot be compiled on their own. I have chosen not to pass them to the `ignore` argument as we still want to see other other valid warnings from these files and we also want them to be checked for formatting.

This resolves #12.


mirenradia:
As described above, it is expected that the lint workflow will fail if `.impl.hpp` files are modified so it is fine to approve this with that check failing, @julianakwan.

julianakwan:
Hi Miren - I was looking through the files that had been changed and I noticed that lint changed some virtual void functions in GRAMR and GRAMRLevel to just void, e.g. post_restart(). Is this because there was nothing further down the inheritance chain that required them to be virtual? I supposed if there was a need for them to be virtual later on, we could change it by hand.

mirenradia:
> Hi Miren - I was looking through the files that had been changed and I noticed that lint changed some virtual void functions in GRAMR and GRAMRLevel to just void, e.g. post_restart(). Is this because there was nothing further down the inheritance chain that required them to be virtual? I supposed if there was a need for them to be virtual later on, we could change it by hand.

Actually, it's still possible to override the virtual function from the base classes (`Amr` and `AmrLevel` respectively) since we're using `override` rather than `final`. The C++ Core Guidelines recommends using one of `virtual`, `override` or `final` for virtual functions (see [here](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override)) hence why clang-tidy removed the redundant `virtual`.

julianakwan (approved):

Issue #40, opened 2024-01-08
Issue with fourth order cell-centered interpolation


mirenradia:
When this code was ported from Chombo to AMReX, Weiqun implemented the following interpolation methods in AMReX in order to improve the accuracy:
* Fourth order cell-*centred* interpolation from coarse to fine (AMReX-Codes/amrex#2960).
* Fourth order interpolation from fine to coarse (AMReX-Codes/amrex#2987).

We believe GRChombo uses the following instead:
* Fourth order cell-*average* interpolation from coarse to fine.
* Simple averaging from fine to coarse (should be second-order accurate).

Unfortunately, these seem to have caused a accuracy regression from GRChombo. The GIFs below show a visualization of the CCZ4 variable $\Theta$ (which is a proxy for the constraint violations in the data and should be damped by the constraint damping) on a slice through the equatorial plane of the first few timesteps of a BBH simulation ([params_grteclyn.txt](https://github.com/GRTLCollaboration/GRTeclyn/files/13862956/params_grteclyn.txt) for GRTeclyn and [params_grchombo.txt](https://github.com/GRTLCollaboration/GRTeclyn/files/13862993/params_grchombo.txt) for GRChombo). Note that the zoom nor the number of visualized timesteps is not consistent between the GIFs.
* GRTeclyn ([commit 1fd5701](https://github.com/GRTLCollaboration/GRTeclyn/tree/1fd5701ce43193f69573db2184b1a277821d22d8)):
  ![GRTeclyn-Theta](https://github.com/GRTLCollaboration/GRTeclyn/assets/32646026/75333e27-583f-454d-ac1e-0506d005418a)
* GRChombo ([commit 72a6d50](https://github.com/GRTLCollaboration/GRChombo/tree/72a6d50354e2c9eab8b6cc275b703874d2c0c877)):
  ![GRChombo-Theta](https://github.com/GRTLCollaboration/GRTeclyn/assets/32646026/41c2f024-d557-4290-bfc9-ea2c03e1040b)

Switching back to the methods analogous to the ones in GRChombo ([commit a8b0e11](https://github.com/GRTLCollaboration/GRTeclyn/commit/a8b0e11e28b19978bade800cc894526d2735f66d)) makes GRTeclyn give similar results to GRChombo:
* GRTeclyn ([commit a8b0e11](https://github.com/GRTLCollaboration/GRTeclyn/tree/a8b0e11e28b19978bade800cc894526d2735f66d)):
[...]

Issue #16, opened 2023-02-13
Port Surface Extraction capabilities


mirenradia:
This depends on #2.

mirenradia:
As I [commented](https://github.com/GRTLCollaboration/GRTeclyn/pull/141#issuecomment-3928928163) on #141, in the PR for this, let's remove the old `AMRInterpolator` folder and rename the following classes/files to remove the word "Particle" (only added to avoid conflicts with the names of things in `AMRInterpolator`:

* `InterpolationLayoutParticle`
* `InterpolationQueryParticle`
* `MPIContextParticle`
* `MPILayoutParticle`

PR #63, opened 2024-05-24
Refactor derived variables

### LINKED ISSUE #41: Refactor handling of derived variables
Currently, (e.g. [in `GRAMRLevel::variableSetUp()`](https://github.com/GRTLCollaboration/GRTeclyn/blob/8b5b6ba9e136b62d3d11ebe5cb72bccce607aee7/Source/GRTeclynCore/GRAMRLevel.cpp#L104-L148)), the derived variables are hardcoded to being the constraint ones (i.e. `Ham`, `Mom1`, `Mom2` and `Mom3`).

Since we don't want to specialise to GR/CCZ4 examples at the level of `GRAMRLevel`, we should move this code to the relevant examples and also add an interface so that it is simple for developers of new examples to add their own derived variables.

We should also remove any references to the GRChombo concept of "diagnostic variables" and replace them with derived variables.

Resolving this issue is a dependency of #2.

mirenradia:
This PR resolves #41 by moving the variable setup and `derive()` function into the specific level class (well just `BinaryBHLevel` for now) rather than `GRAMRLevel`.

It also does the following
* Adds `GRAMRLevel::stateVariableSetUp()`. This needs to be called by `PhysicsLevel::variableSetUp()`. We could instead not require this and call `stateVariableSetUp()` directly in `DefaultLevelFactory::variableSetUp()`?
* Adds Weyl4
  * Adds a test which computes this on the random test initial data, outputs to a HDF5 file and then compares it with an HDF5 file produced by the analagous test from GRChombo. The test is skipped when not built with HDF5. I have added HDF5 to the unit tests GH action workflow
* Removes some old unused code paths guarded by `AMREX_USE_HDF5` since we now use HDF5.
* Adds the ability for `derive()` to calculate derived quantites on multifabs with ghost cells (which we will need for #2). Previously, for GRChombo we used to fill these ghost cells (using BCs as appropriate) but this would have been very complicated so now we just fill a source multifab with extra ghost cells.
* Switches to using hardcoded parities rather than letting them be parameters and removes the `vars_parity` and `vars_parity_diagnostic` parameters.
* Renames the variable types as follows for consistency with AMReX:
  * evolution -> state
  * diagnostic -> derived
* Removes the notion of fixed diagnostic variables since derived variables are instead defined in the examples `PhysicsLevel::variableSetUp()` function. MultiFabs are instantiated as
required to store them and the component numbers are not fixed.
* Renames `UserVariables` to `StateVariables` again for consistency with AMReX.
* Adds a test for the Constraints similar to the Weyl4 one.
* Removes the old `Constraints.[impl.].hpp` and renames `NewConstraints.hpp` to `Constraints.hpp`. The former was only kept in GRChombo for backwards compatibility reasons.

I still need to make some changes to resolve some of the clang-tidy warnings but, given the size of this PR, I thought I would create it now so that you can start reviewing next week, @julianakwan.


Examples/BinaryBH/BinaryBHLevel.cpp
julianakwan, line 1: The derived variables are cell_quartic but the state variables are quartic?

Examples/KerrBH/KerrBHLevel.hpp
julianakwan, line 1: I think it's meant to be initData() in AMReX?

julianakwan: Probably just needs to be AMReX-ified.

Source/utils/WeylExtraction.hpp
julianakwan, line 1: Is this removed because WeylExtraction hasn't been ported?

Examples/BinaryBH/BinaryBHLevel.cpp
mirenradia: Perhaps I should move the change of the state variables to `quartic` to a separate PR so that we can explicitly link it to #40. I guess we can leave the derived variables as `cell_quartic` since they don't affect the evolution but maybe we should be consistent.

Examples/KerrBH/KerrBHLevel.hpp
mirenradia: Yeah, this example hasn't been ported but it made sense to remove all codepaths guarded by `AMREX_USE_HDF5` in 89913aa4f936926720403b9b7ade2480d22f42c8.

Source/utils/WeylExtraction.hpp
mirenradia: Yes, this was just to make it compile. This will be fixed when #16 is done.

julianakwan (changes requested):
Nice - the proposed changes work well! I agree that the stateVariableSetUp should be a part of GRAMRLevel that each PhysicsLevel::variableSetUp can call, since not all applications need the extra CCZ4/BSSN variables. 
I tested this pull request by merging this branch with the scalar field example that I'm working on. I found a few minor quibbles (so I'll mark it as 'request changes'. Note that I would be happy to leave these as future work on a separate pull request if you would like to merge this branch into develop immediately.)

-  It's a bit cumbersome to have to set the parity for all variables even though reflective boundary conditions are not used. It would be nice if there was a separate option that checked for the type of BCs and then if some of them are reflective, look for the parities. 
- Consider changing the definition of variables as enums to be a vector of strings like what AMReX requires when interfacing with plotMultiFab, derive.lst etc. I find myself making a separate vector of strings that just goes over the enums so I can pass variable names to AMReX functions. 
- I think some of the code in PhysicsLevel::derive() can be placed in GRAMRLevel and then called by PhysicsLevel like stateVariableSetUp. For example the Hamiltonian and momentum calculations would be common to any GR application and so will be copied in each GR physics example and this is probably a point of failure (if we update the calculations later on).

mirenradia:
> Nice - the proposed changes work well! I agree that the stateVariableSetUp should be a part of GRAMRLevel that each PhysicsLevel::variableSetUp can call, since not all applications need the extra CCZ4/BSSN variables. 

I guess my question was whether we wanted to make `DefaultLevelFactory::variableSetUp()` call this so there would be no need to define a `PhysicsLevel::variableSetUp()` if no non-state variables were required?

> * It's a bit cumbersome to have to set the parity for all variables even though reflective boundary conditions are not used. It would be nice if there was a separate option that checked for the type of BCs and then if some of them are reflective, look for the parities.

This is a good idea. Let me look at setting some kind of default so that the user doesn't need to set this if they never intend to use reflective BCs.

> * Consider changing the definition of variables as enums to be a vector of strings like what AMReX requires when interfacing with plotMultiFab, derive.lst etc. I find myself making a separate vector of strings that just goes over the enums so I can pass variable names to AMReX functions.

I presume you mean `StateVariables::names` which is currently of type `std::array<std::string, NUM_VARS>`? We went for `std::array` at the time for GRChombo as the size is known at compile time but since this isn't really performance critical, I guess there isn't much of a downside to switching to `amrex::Vector<std::string>`.

> * I think some of the code in PhysicsLevel::derive() can be placed in GRAMRLevel and then called by PhysicsLevel like stateVariableSetUp. For example the Hamiltonian and momentum calculations would be common to any GR application and so will be copied in each GR physics example and this is probably a point of failure (if we update the calculations later on).

We don't want to include the constraint code in `GRAMRLevel` since it's GR specific and `GRAMRLevel` should be used for non-GR examples too. However, I guess some of the setup code that generates the source multifab can be moved into a `GRAMRLevel` function.

julianakwan:
I also forgot to mention: some of the 'old' files in the test directory e.g. `Tests/CCZ4RHSTest/ADMConformalVars-fdf5a7a.hpp` have been modified but are tagged with a SHA from before the refactoring of the derived variables. Just commenting here so that we remember to update these with the new SHA once this pull request has been merged into the `devel' branch.

mirenradia:
> I also forgot to mention: some of the 'old' files in the test directory e.g. `Tests/CCZ4RHSTest/ADMConformalVars-fdf5a7a.hpp` have been modified but are tagged with a SHA from before the refactoring of the derived variables. Just commenting here so that we remember to update these with the new SHA once this pull request has been merged into the `devel' branch.

I just had a look at the changes in these files and it's just coming from the renaming `UserVariables` -> `StateVariables`. I don't think we should rename them and even if we were, I have no idea what we would rename them to.

Examples/BinaryBH/BinaryBHLevel.cpp
mirenradia: This is resolved now that #69 is merged and I have rebased this branch. All variables use `cell_quartic` interpolation.

mirenradia:
Following 8bd6e68334866db4c3bcdd0048643267f2658136, this PR now depends on #70 so that should be merged first. This branch is based off the branch in that PR.

mirenradia:
I should add that 8bd6e68334866db4c3bcdd0048643267f2658136 means that we now use `AmrLevel::derive()` rather than using our own override. This is possible because AMReX-Codes/amrex#3990 means that we can use the new MultiFab version of `DeriveFunc` which allows us to use the new, more efficient MultiFab ParallelFor.

julianakwan (changes requested):
Thanks so much for changing the variable names to be vectors - I think that was a really handy addition. Also the new `set_up` and `compute_mf` functions are nice. 

I ran into a few issues running the BinaryBH example though: 
1. I can't seem to get the derived variables to print out in the plot files. I'm using `inputs.test` as the parameter file which contains the lines:
```
amr.plot_vars = chi # The default is to plot all state variables
amr.derive_plot_vars = Ham Mom # The default is none for derived variables.`
```
I also added `calculate_constraint_norms = true` 

I think it would be nice if `inputs.test` (or some other input file) could be modified to contained a working example with the derived variables outputted.  

2. A separate issue is if I do `amr.derive_plot_vars = ALL` then the code exits with a segmentation fault when writing the plot files. It happens in the Weyl4 calculation (I've saved the Backtrace.* files if you want to see?) I think it is because the default value for `calculate_constraint_norms` is false but `ALL` makes it calculate everything anyway and there are now too many components in `deriv_lst` - if I set `calculate_constraint_norms = true` and `amr.derive_plot_vars = ALL` then I get every derived variable except `Mom`. 

I think this could be fixed by adding an `if` statement to `SimulationParameters.hpp` to check if `ALL` is set then automatically set `calculate_constraint_norms = true` as well. I'm not sure about the missing `Mom` though...

mirenradia:
> I ran into a few issues running the BinaryBH example though:
> 
>     1. I can't seem to get the derived variables to print out in the plot files. I'm using `inputs.test` as the parameter file which contains the lines:
> 
> 
> ```
> amr.plot_vars = chi # The default is to plot all state variables
> amr.derive_plot_vars = Ham Mom # The default is none for derived variables.`
> ```

For this parameter, you need to use the `name` passed to the derive list. In this case, the name is `constraints` which contains 4 components. I agree this is a bit confusing. We should document it somewhere.

> I also added `calculate_constraint_norms = true`

This parameter currently doesn't do anything. Previously, in GRChombo it would calculate the L2 norm of the constraint variables and write them to a file (see [here](https://github.com/GRTLCollaboration/GRChombo/blob/main/Examples/BinaryBH/BinaryBHLevel.cpp#L173-L193)). We can now re-implement this but given how long it has taken to merge this PR, I suggest creating an issue (about AMR reductions) and deferring it to a later PR.

I presume you thought this would change what is passed to the `a_calc_mom_norm` argument in `Constraints::set_up()`? In GRChombo, it was impossible to change it at runtime as the number of diagnostic variables was hardcoded so it was intended to be hardcoded at the example level. We could make it a parameter but I don't think it's worth it.

>
> I think it would be nice if `inputs.test` (or some other input file) could be modified to contained a working example with the derived variables outputted.
>

I already added `amr.derive_plot_vars = ALL` to `params_test.txt` but perhaps it would be worth removing these lines from `inputs.test` to avoid confusion.

> 2. A separate issue is if I do `amr.derive_plot_vars = ALL` then the code exits with a segmentation fault when writing the plot files. It happens in the Weyl4 calculation (I've saved the Backtrace.* files if you want to see?) I think it is because the default value for `calculate_constraint_norms` is false but `ALL` makes it calculate everything anyway and there are now too many components in `deriv_lst` - if I set `calculate_constraint_norms = true` and `amr.derive_plot_vars = ALL` then I get every derived variable except `Mom`.
> 
> 
> I think this could be fixed by adding an `if` statement to `SimulationParameters.hpp` to check if `ALL` is set then automatically set `calculate_constraint_norms = true` as well. I'm not sure about the missing `Mom` though...

I was able to reproduce this. I think the problem comes from saying that `Weyl4` only depends on all state variables up to and excluding `c_B1` [here](https://github.com/GRTLCollaboration/GRTeclyn/blob/ca79b14a9660bc6fba6c42a17f26a5800bea73b7/Source/CCZ4/Weyl4.impl.hpp#L342) but then tries to load all of the CCZ4 variables [later](https://github.com/GRTLCollaboration/GRTeclyn/blob/ca79b14a9660bc6fba6c42a17f26a5800bea73b7/Source/CCZ4/Weyl4.impl.hpp#L20). Weird that the regression test didn't pick this up. Maybe we should modify it to build with `DEBUG = TRUE` (or create a matrix like for the unit tests).

julianakwan (approved):
Thanks! I did a fresh pull from the repo and confirmed that my problems had been fixed. 

> I presume you thought this would change what is passed to the a_calc_mom_norm argument in Constraints::set_up()? 

Yeah, this is exactly what I thought was going on! I agree implementing `calculate_constraint_norms` should be a separate issue. 

Thanks for updating `inputs.test` also - I think some other parameter files might also have the "old" style of specifying constraint parameters but this should be a separate issue also.

PR #71, opened 2024-07-19
Build: Refactor makefiles

### LINKED ISSUE #68: Potential race condition in base GNUmakefile examples target
@WeiqunZhang pointed out that the `examples` target in the base `GNUmakefile` has a potential race condition since all of the individual examples targets will try to build AMReX source files together at the same time if built with parallel jobs.

This is why your CI keeps failing for the ScalarField, @julianakwan!

mirenradia:
This adds a base Makefile that should be common to GRTeclyn examples (once there is more than one) and eventually the tests too.

There is also a new `amrex-objects` target in the root makefile which should fix #68 as it is a dependency of the examples targets although this does rely on the examples keeping the same default configuration (i.e. `USE_MPI = TRUE`, `COMP = gnu`).

I have added a base name to the BinaryBH example.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -14,6 +14,9 @@ jobs:
       AMREX_HOME: ${{ github.workspace }}/amrex
       BINARYBH_EXAMPLE_DIR: ${{ github.workspace }}/GRTeclyn/Examples/BinaryBH
       GCC_VERSION: 14
+      TMP_BUILD_DIR: ${{ github.workspace }}/GRTeclyn/tmp_build_dir
+      EXECUTABLE: BinaryBH3d.gnu.MPI.ex
+      PARAMS: params_test.txt
 
     steps:
     - name: Checkout AMReX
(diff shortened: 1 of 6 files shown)


mirenradia:
@julianakwan, would you be able to check this resolves the CI issues you were finding on your matter/scalar field branches?

julianakwan:
> @julianakwan, would you be able to check this resolves the CI issues you were finding on your matter/scalar field branches?

Yeah, I'll test it out as a part of my usual suite of checks for PRs :)

julianakwan:
I merged the `bugfix/68` branch into the `enhancement/scalar_fields` branch to test if the new Makefile would fix the race condition in the Example builds but unfortunately it did not - [see failed GPU workflow here](https://github.com/GRTLCollaboration/GRTeclyn/actions/runs/10010928964/job/27673240152#step:8:12062). 
@mirenradia suggested separating out each example in the `Examples` directory and building them individually as a possible workaround.

mirenradia:
This PR is incompatible with #82 so I'm going to close it in favour of that.

PR #74, opened 2024-09-26
CI: Fix ROCm and oneAPI versions

julianakwan:
The HIP build of the Binary BH example was failing the GPU build workflow because of some missing ROCm dependencies so I changed the location of repo it would install from. However, this does fix the version of ROCm to 6.2.1

### DIFF
--- a/.github/workflows/gpu-build.yml
+++ b/.github/workflows/gpu-build.yml
@@ -30,9 +30,9 @@ jobs:
       BUILD_ARGS: USE_${{ matrix.gpu-backend }}=TRUE DEBUG=${{ matrix.debug }} USE_MPI=TRUE
       CUDA_MAJOR_VERSION: 12
       CUDA_MINOR_VERSION: 0
-      ROCM_VERSION: 6.2.1
+      ROCM_VERSION: 6.2.2
       ONEAPI_COMP_VERSION: 2024.2
-      ONEAPI_MPI_VERSION: 2021.9.0
+      ONEAPI_MPI_VERSION: 2021.13
 
     steps:
     - name: Checkout AMReX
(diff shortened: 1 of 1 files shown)


mirenradia:
I re-ran the workflow on the tip of the `develop` branch (089ad0e926de1be713dbeec8aa6b5d3bc6cd44c8) and this doesn't seem to be a problem any more so I would be happy to close this PR without merging.

If we want to fix the version of ROCm, I would be OK with that but, for consistency with what we do for CUDA, maybe move the version to a workflow environment variable e.g. `ROCM_VERSION` (I don't think there's a need to separate by major, minor and patch versions like for CUDA) which is evaluated where needed.

julianakwan:
After some discussion in the GRTeclyn telecon, I fixed the version of ROCm to 6.2.1, the DPC++ compiler to 2024.2 and the oneAPI MPI library to 2021.9.0 which are the latest versions (I guess they don't release Intel MPI very often?)

.github/workflows/gpu-build.yml
```diff
@@ -30,6 +30,9 @@ jobs:
       BUILD_ARGS: USE_${{ matrix.gpu-backend }}=TRUE DEBUG=${{ matrix.debug }} USE_MPI=TRUE
       CUDA_MAJOR_VERSION: 12
       CUDA_MINOR_VERSION: 0
+      ROCM_VERSION: 6.2.1
+      ONEAPI_COMP_VERSION: 2024.2
+      ONEAPI_MPI_VERSION: 2021.9.0
```
mirenradia: I think the latest currently available versions are as below. Looking at the APT package names, I think the patch level is omitted for the Intel oneAPI packages (the latest MPI version is 2021.13.1).
```suggestion
      ROCM_VERSION: 6.2.2
      ONEAPI_COMP_VERSION: 2024.2
      ONEAPI_MPI_VERSION: 2021.13
```

mirenradia (changes requested):
Thanks for the PR. Just the 1 small change below.

mirenradia (approved):

PR #82, opened 2024-12-20
Build: Remove default `TMP_BUILD_DIR`

mirenradia:
I think this will help with #68. The disadvantage (and the reason I bothered to set this default in the first place) is that consecutive builds of different AMReX executables (e.g. two different examples) will be slower. To mitigate this, users should use Ccache which I have added to the README. I will make a separate PR to add Ccache to the CI workflows.

### DIFF
--- a/Tools/GNUMake/Make.defaults
+++ b/Tools/GNUMake/Make.defaults
@@ -24,8 +24,3 @@ BL_NO_FORT = TRUE
 AMREX_NO_PROBINIT = TRUE
 
 CXXSTD = c++17
-
-GRTECLYN_HOME ?= $(realpath ../..)
-
-# Use a common temporary build directory for all GRTeclyn applications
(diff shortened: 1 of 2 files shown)


Tools/GNUMake/Make.defaults
julianakwan, line 1: Why has the definition of GRTECLYN_HOME been removed? I think people should be defining it when they set up their environment in their .bashrc etc. but I thought it was nice catch in case they forgot...?

mirenradia: In this case, the variable was only being used by 1 line in this file which was removed.

There shouldn't be any need for users to define this themselves as it's only used internally.

julianakwan (approved):
Thanks for making these changes. I've checked that this resolves the race condition when there is more than one example being built in the `Examples` directory

PR #84, opened 2025-01-09
AMRInterpolator: Use the correct name (m_dx) for the member variable dx

julianakwan:
This is a quick bug fix - Intel oneAPI 2025.0.4 compilers will complain about a misnamed member variable in `Lagrange.impl.hpp` I'm not sure why it wasn't picked up before!


mirenradia (approved):
I'm also not sure why this problem only appeared just now and older compilers missed it. Looks like it's Clang/LLVM 19 which first picked it up. I think the 2025.0 version of the oneAPI DPC++ compiler is based on this.

PR #85, opened 2025-01-17
Port fixed grids tagging

julianakwan:
This ports the fixed tagging criterion in which the inner L/4 of the box is tagged for refinement to be compatible with AMReX. This is useful for weak scaling tests and also examples in which chi is not needed.

### DIFF
--- a/Source/TaggingCriteria/FixedGridsTaggingCriterion.hpp
+++ b/Source/TaggingCriteria/FixedGridsTaggingCriterion.hpp
@@ -25,22 +25,27 @@ class FixedGridsTaggingCriterion
         const std::array<double, AMREX_SPACEDIM> a_center)
         : m_dx(dx), m_L(a_L), m_level(a_level), m_center(a_center){};
 
-    template <class data_t> void compute(Cell<data_t> current_cell) const
+    template <class data_t>
+    AMREX_GPU_DEVICE data_t
+    compute(int i, int j, int k,
+            const amrex::Array4<const data_t> &current_cell) const
     {
         data_t criterion = 0.0;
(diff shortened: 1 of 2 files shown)


Source/TaggingCriteria/FixedGridsTaggingCriterion.hpp
```diff
@@ -14,33 +14,39 @@
...
         const double dx, const int a_level, const double a_L,
         const std::array<double, AMREX_SPACEDIM> a_center)
         : m_dx(dx), m_L(a_L), m_level(a_level), m_center(a_center){};
-
-    template <class data_t> void compute(Cell<data_t> current_cell) const
+    // NOLINTEND(bugprone-easily-swappable-parameters)
+    template <class data_t>
+    AMREX_GPU_DEVICE data_t
+    compute(int i, int j, int k,
+            const amrex::Array4<const data_t> &current_cell) const
```
mirenradia: I don't think the `current_cell` argument is used so it should be removed.
```suggestion
    compute(int i, int j, int k) const
```

mirenradia (changes requested):
Thanks for the PR. Just 1 minor change.

I was debating whether it's worth replacing the `simd_xxx` functions with more conventional ones but this can be done later if and when we decide to remove the SIMD infrastructure.

Source/TaggingCriteria/FixedGridsTaggingCriterion.hpp
julianakwan: Thanks! I've removed `current_cell' as suggested.

mirenradia (approved):
LGTM

Issue #17, opened 2023-02-13
Add basic documentation


mirenradia:
Consider switching to [Sphinx](https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html) rather than using the GitHub wiki.

KAClough:
I have considered this, and I would be happy to stick with the GitHub wiki, which has so far worked fine for us :-)

KAClough:
@mirenradia makes a good point about being able to update the wiki when doing a PR for an update to the code. A solution that is suggested is to add the GitHub wiki as a submodule of the code... but it seems a bit complicated...

mirenradia:
I don't think that would work very well. I think it's best for the documentation source files (e.g. markdown files) to just reside in the code repository (e.g. in a docs subdirectory) and then it can be built using CI/CD.

KAClough:
I would like to keep the ability to just modify the docs using the GUI in GitHub though. Ideally the system could support both. Making it something that requires CI/a commit and PR every time creates a barrier to people (me) making small but useful amendments.

mirenradia:
You can edit Markdown files on GitHub but yes you would need to open a PR to make a change. I agree that it is a little bit more of a barrier but I don't think this is a bad thing for the following reasons:

1. If it's just a minor update, you wouldn't need to put much effort into the PR description or title. It's only a minor barrier.
2. A small update to the documentation PR can be an easy way for a new user to get into the habit of contributing to the code.
3. Since PRs are reviewed, the quality of the documentation may be improved.

KAClough:
This has been started in the wiki here, but we have decided to move it to the read the docs format but using markdown.

julianakwan:
We have a [documentation page](https://grtlcollaboration.github.io/GRTeclyn/) now! I've made a start on generating online documentation with `MkDocs`. The markdown files are kept in the [enhancement/documentation](https://github.com/GRTLCollaboration/GRTeclyn/tree/enhancement/documentation) branch in the `./docs` directory and I build to the `gh-pages` branch. I've moved (some of) the wiki pages over but it is still very much a work in progress. NB: I've had to change the branch where GitHub pages loads the site information from `develop` to `gh-pages`.

Please feel free to edit or add to the pages. Building the docs is done manually at the moment (but I will add it to the CI/CD at some point). Here is a quick guide on how: 

0. Check out the `enhancement/documentation` branch. 
[...]

PR #86, opened 2025-01-24
Refactor tagging

mirenradia:
This PR refactors the code that tag cells for refinement. There are 3 main changes which can be found in the commit messages but I'll copy them here for completeness

## [Tagging: Remove template parameter from `ChiExtractionTaggingCriterion`](https://github.com/GRTLCollaboration/GRTeclyn/commit/b4e5bf7f61e4bcddbf06d354a5b4c7968461e876)

To do this, we instead change the types of `extraction_radii` and `extraction_levels` (in the `SphericalExtraction::params_t` struct) to `amrex::Gpu::ManagedVector`s and pass pointers to their underlying data to the tagging criterion compute class.

## [Core: Refactor `GRAMRLevel::errorEst`](https://github.com/GRTLCollaboration/GRTeclyn/commit/a143b0fcad2dbbf92385b3e3ffeab7b98e8b651a)

This function is a pure virtual one in `amrex::AmrLevel` but it has a lot of redundant arguments which we don't use. For simplicity, it now calls two virtual functions with fewer arguments.

* `pre_tag_cells()`
* `tag_cells()`

which can be implemented in the example's derived class. The former is optional and the latter is mandatory (pure virtual).

## [Tagging: Generalize tagging criteria](https://github.com/GRTLCollaboration/GRTeclyn/commit/d0de7e4a42c4a6f95e71005c6960a88731f55220)

Rather than requiring the use of a tagging criterion which calculates a value for each cell and then forcing the tagging of cells where the value exceeds the `regrid_threshold`, we now allow for "taggers" which explicitly tag cells as they wish. This is a bit more flexible and it avoids having to use a large magic number (e.g. 100.0) to ensure cells we want to be tagged are definitely tagged independent of the `regrid_threshold`.

I have converted `FixedGridsTaggingCriterion` to `FixedGridsTagger` in the final commit (09e4f19258ffab1b3f218cc23db30ff8cc84ccb0).

I realise the linting CI will fail so will open as a draft for now and fix it later.


Examples/BinaryBH/BinaryBHLevel.cpp
julianakwan, line 1: Is there a reason you got rid of `simpar` as a shortcut for `simParams()`? I think it just reads better without the braces, but pretty minor.

Source/GRTeclynCore/GRAMRLevel.hpp
julianakwan, line 1: I think we need some documentation on this. At first I thought `pre_tag_cells` was to prepare the `TagBoxArray` containing the tags by e.g. resetting the values of `amrex::TagBox::SET` again. But now I see that is for callling `FillPatch`. I see that you have comments but it is in the `GRAMRLevel` files not the physics level files.

julianakwan (changes requested):
Thanks for refractoring these! It makes sense to me and looks good to merge once the lint issues have been fixed. 
I see the `simd` functions have been removed from the `FixedGridTagger`.

Examples/BinaryBH/BinaryBHLevel.cpp
mirenradia: I just thought it was an unnecessary line and didn't add anything. This will probably be changed anyway when we refactor parameter loading.

Source/GRTeclynCore/GRAMRLevel.hpp
mirenradia: There is a comment above the declaration in BinaryBHLevel.hpp, just a bit shorter than that in GRAMRLevel.hpp but I thought the definition in BinaryBHLevel.cpp and the comment in the virtual declaration in GRAMRLevel.hpp would be enough.

I agree that we should add more comprehensive documentation but this is probably part of #17.

mirenradia:
> I see the `simd` functions have been removed from the `FixedGridTagger`.

I didn't see the point in keeping them given the loop wouldn't have worked with SIMD anyway.

julianakwan (approved):
Thanks for fixing those annoying lint errors!

mirenradia:
> Thanks for fixing those annoying lint errors!

I think a lot of the "fixes" were rather adding files to ".lint-ignore"!

PR #88, opened 2025-02-05
CI/Format: Update lint action and clang-format to v19

mirenradia:
This PR updates the lint CI workflow and git pre-commit configuration to use the latest version (19.1.17) of the LLVM tools clang-tidy and clang-format.

I have also applied formatting changes and fixes/ignores for the linter warnings that arise from this update.


julianakwan (approved):
Thank you! Some of those formatting changes brought about by the update to clang-format could also be caught by the local editor rather than the pre-commit hook e.g. I have a special emacs command to apply clang formatting as I write - does this mean everyone's clang-format.el file (for example) should be updated as well?

mirenradia:
>  I have a special emacs command to apply clang formatting as I write - does this mean everyone's clang-format.el file (for example) should be updated as well?

I think this probably uses the version of `clang-format` available on your system so you may need to update it.

julianakwan:
Yeah I think so too - we should update the wiki with the correct version as well. I realise that this is technically the GRChombo wiki, which is a bit terrible because there is no way I can remember all this stuff!

PR #89, opened 2025-02-07
Black Holes: Port puncture tracking and enable tagging around the punctures

### LINKED ISSUE #15: Port puncture tracker
This depends on #2.

mirenradia:
This long-awaited PR ports the puncture tracking capabilities from GRChombo to GRTeclyn. Here, we leverage AMReX's particle capabilities and the particles are moved around the grid as particles. This simplifies working out what box/grid they reside in as we can instead just make AMReX redistribute them.

The main things introduced/changed by this PR are described below

## Parameters

The parameters have been renamed to

* `puncture_tracking.enabled` = `false` (default) or `true`
* `puncture_tracking.level` = `0` (default), ..., `max_level`

There are also some new ones, some of which are only used in the BinaryBH example:

* `puncture_tracking.writeout_level` = `0` (default), ..., `max_level`
* `puncture_tracking.disable_writeout` = `false` (default) or `true`
* `puncture_tracking.filename` = `'punctures'` (default) or another string
* `puncture_tracking.output_path` = `.` (default)

We might want to change the last one to a GRTeclyn-wide `output_path` parameter at some point.

## `PunctureTracker` class

This is now derived from `amrex::ParticleContainer<AMREX_SPACEDIM, 0>` and is templated over `num_punctures` (which will probably be `2`). The member functions have been renamed from the ones in `PunctureTracking` in GRChombo.

For the actual tracking, we essentially copy what AMReX's [`TracerParticleContainer::AdvectWithUcc`](https://github.com/AMReX-Codes/amrex/blob/823ec7fc00a9fc43b95f4b7f86b376840682d072/Src/Particle/AMReX_TracerParticles.cpp#L120-#L194) does except we advect in the reverse direction to the shift. This uses the midpoint method with linear interpolation (which means we only need to fill 1 ghost cell).

Note that we also store the coordinates in `PunctureTracker::m_puncture_coords` variable which is set and broadcast to all MPI processes in `PunctureTracker::update_puncture_coords`.

## `BHAMR` changes

This is now also templated over `num_punctures` so it can store a puncture tracker with the correct number of punctures! Its constructor initialises the `PunctureTracker` if `puncture_tracking.enabled == true`.

## New `GRAMRLevel` hooks

I have added some hooks for examples to do things post init, post restart, pre checkpoint and post checkpoint. I should make the formatting of these hooks more consistent...

## `PunctureTagger` class

Rather than having a combined `ChiPunctureExtractionTaggingCriterion` class as before, we switch to using the existing `ChiExtractionTagger` class and then using a separate `PunctureTracker` class which just does the tagging around the punctures.

This resolves #15.

I'll open this as a draft for now as I still need to make a few minor changes and fix the linter errors but I think it should be close enough to finish for you to test, @KAClough.


KAClough:
This all looks good. I am still testing different cases, but the puncture tracking currently seems to fail on regridding - I am running the params_profile.txt case (with puncture tracking enabled), and when it gets to t=1.25 level 3 regrids and the puncture.dat file locations all go to zero.

mirenradia:
> This all looks good. I am still testing different cases, but the puncture tracking currently seems to fail on regridding - I am running the params_profile.txt case (with puncture tracking enabled), and when it gets to t=1.25 level 3 regrids and the puncture.dat file locations all go to zero.

I will try to reproduce and investigate.

mirenradia:
I think the problem is, because we're using reflective BCs, the puncture particles are just leaving the valid domain (since they're at $z = 0$) and then AMReX is marking them as invalid. I will work on a fix for this problem but in the mean time you should be able to try without reflective BCs.

KAClough:
> I think the problem is, because we're using reflective BCs, the puncture particles are just leaving the valid domain (since they're at z = 0 ) and then AMReX is marking them as invalid. I will work on a fix for this problem but in the mean time you should be able to try without reflective BCs.

Yes, I can confirm that this works ok without the reflective BCs, and it also restarts ok from a checkpoint (with the correct puncture values matching the dat file). In the punctures.dat file it doesn't delete the duplicate time data like we used to using the SmallDataIO function, but this is easy to fix. 

I will try it on GPUs today and also make a hacky 3 punctures example to see if that works.

mirenradia:
> Yes, I can confirm that this works ok without the reflective BCs, and it also restarts ok from a checkpoint (with the correct puncture values matching the dat file). In the punctures.dat file it doesn't delete the duplicate time data like we used to using the SmallDataIO function, but this is easy to fix.

I think I've resolved both of these problems now. Reflective BCs should work (I've clamped the punctures to stay within the domain) and so should the removal of duplicate time data in the `punctures.dat` file.

KAClough:
I will test again this updated version, but the hacky triple BH I made works fine, and copes with the BHs merging (we still retain the 3 punctures, but that's ok). Here is a nice picture:

image

mirenradia:
## Puncture Tracker test

I have added a new test which sets the y-component of the shift to a constant uniform value and then tracks some "punctures". Since puncture tracking is just advection in the reverse direction to the shift, we know analytically how they should move. This test also further checks the puncture tagger by ensuring the finest 2 levels cover spheres around the punctures with radius that is a function of the "mass".

KAClough:
This looks good to me! The following has been checked:
- The code compiles and runs on Cosma8 CPUs
- The code works with reflective boundaries
- The code can restart from checkpoint
- The output of the punctures works ok including on restart
- A 3BH example works ok too

It is still queuing to run on the MI300x GPUs but does compile on them so it is probably ok.

UPDATE: It did run fine on the GPUs so it all looks good to me.

Examples/BinaryBH/inputs.test
julianakwan, line 1: What was the function of this input file? I realize that `params_test.txt` is needed for the regression test, but if it is for other testing purposes, it could be replaced with `params_very_cheap.txt`. I know this is not the point of this PR but I just think our input files are a bit messy.

Examples/BinaryBH/BinaryBHLevel.cpp
julianakwan, line 1: In `set_puncture_coords`, could there be some sort of check implemented for the number of punctures? It does assume two and you've nicely templated over the number of punctures already, so it's not hard to imagine that someone might add extra punctures...? You could use`num_punctures` to enforce the expected number.  

Also this seems inconsistent with `initial_coords` as a parameter?

julianakwan: Ah I see that `initial_coords` is only in the unit test.

Tests/Common/doctestCLIArgs.hpp
julianakwan, line 1: I don't quite understand when this `insert` function is needed? Maybe some more comments would be nice...

Tests/PunctureTrackerTest/Make.package
julianakwan, line 1: Please reorder headers alphabetically.

Tests/PunctureTrackerTest/PunctureTrackerTest.cpp
julianakwan, line 1: Maybe all our headers should be ordered alphabetically? (I need to do this too :) )

Tests/PunctureTrackerTest/PunctureTrackerLevel.cpp
julianakwan, line 1: Headers could be reordered alphabetically here too.

Examples/BinaryBH/BinaryBHLevel.cpp
julianakwan, line 1: I think the punctures should only be written at every coarse time step so maybe uncomment out L303-304? It's a lot of output otherwise. It could also be set to be optional if `verbose=1` for example.

julianakwan, line 1: The headers could also be reordered alphabetically.

julianakwan, line 1: I think there should be an option in the params file to not write out the punctures in case of e.g. profiling.

Examples/BinaryBH/inputs.test
mirenradia: I think Weiqun added it.

Yeah, I agree that we should tidy up the parameter files but not in this PR. I'll make an issue for it.

mirenradia: #107

Examples/BinaryBH/BinaryBHLevel.cpp
mirenradia: Clang-format does this but the line
```cpp
// xxxxx #include "SixthOrderDerivatives.hpp"
```
must have messed it up. I'll fix it manually.

mirenradia: Yeah this is a reasonable suggestion. I think it will require reimplementing
`GRAMRLevel::at_level_timestep_multiple` as I think this was removed when the
code was ported to AMReX.

I don't think it should depend on `verbose` since that should only control the
output to stdout.

mirenradia: I'm a bit confused. The `a_puncture_coords` argument in `PunctureTracker<num_punctures>::set_puncture_coords()`
is of type `amrex::Array<amrex::Real, num_puncture_coords>` so necessarily you
can't pass a different number of puncture coords.

Tests/Common/doctestCLIArgs.hpp
mirenradia: Since we pass `argc` and `argv` to `amrex::Initialize()`, it's to add extra
arguments to these. We use this to add the parameter file.

I'll add a comment.

Tests/PunctureTrackerTest/PunctureTrackerLevel.cpp
mirenradia: In their groups, I think they are. I'll add a comment for the different groups
like in `PunctureTrackerTest.cpp`.

Tests/PunctureTrackerTest/PunctureTrackerTest.cpp
mirenradia: I would prefer to keep them in groups like this and clang-format
alphabetizes them within groups.

Examples/BinaryBH/BinaryBHLevel.cpp
julianakwan: We already discussed this in the telecon but I was worried about people changing `num_punctures` but not changing the number of puncture coords used to initialize the puncture tracker. My point is it would be nice if you only had to make the change in one location.

Source/BlackHoles/PunctureTracker.impl.hpp
```diff
@@ -0,0 +1,404 @@
...
+
+    std::string filename{"punctures"}; // default
+    puncture_tracking_pp.queryAdd("filename", filename);
+
+    std::string output_path{"."}; // default
+    // Maybe we might want to change this to a more generic GRTeclyn output_path
+    // at some point
+    puncture_tracking_pp.queryAdd("output_path", output_path);
+
+    if (!FilesystemTools::directory_exists(output_path))
```
julianakwan, line 41: I know it's up to the user to check but you could also add a check if `punctures.dat` exists then move to `punctures.old.<XXXXX>.dat` as in the plot and checkpoint files.

julianakwan (changes requested):
Thanks very much for completing the puncture tracker! I think it looks it great, I especially liked the templating over the number of punctures. I was also a fan of the unit test even if Katy was not as impressed! 

General comments: 

- Please consider trying to add the punctures to the regression test. Perhaps you could try to check the particle locations using `particle_compare.cpp` in the `amrex/Tools/Postprocessing/C_Src` although I do not know if it allows for any error tolerance.
- When running the BinaryBH example on `params.txt` on the PVCs (12 MPI ranks to 12 GPUs), the run stops at (coarse) timestep = 57 because of NaNs in `specificAdvance`:
```
[Level 0 step 57] average evolution speed = 184.1617142 code units/h
[Level 1 step 113] average evolution speed = 184.1555671 code units/h
[Level 2 step 225] average evolution speed = 184.1514987 code units/h
[Level 3 step 449] average evolution speed = 184.1498916 code units/h
[Level 4 step 897] average evolution speed = 184.1485283 code units/h
amrex::Abort::1::NaN in specificAdvance !!!
SIGABRT
amrex::Abort::3::NaN in specificAdvance !!!
SIGABRT
amrex::Abort::6::NaN in specificAdvance !!!
```
I have attached my `params.txt` and `Backtrace.1.txt` (which just repeats the above). The only other thing that I changed was the frequency of the outputs to `punctures.txt`. I'm trying to reproduce it on other machines but I keep getting stuck on the queues - I will update the comments if I get any further. 

[params.txt](https://github.com/user-attachments/files/19346610/params.txt)
[Backtrace.1.txt](https://github.com/user-attachments/files/19346617/Backtrace.1.txt)

**Update 21/03/25**: I just wanted to report that I ran the BinaryBH example on `develop` (same set up as above) and got the same error (but this time at timestep 54) so it seems unrelated:
```
[Level 0 step 54] average evolution speed = 408.2414332 code units/h
[Level 1 step 107] average evolution speed = 408.2205231 code units/h
[Level 2 step 213] average evolution speed = 408.2047277 code units/h
[Level 3 step 425] average evolution speed = 408.1965414 code units/h
[Level 4 step 849] average evolution speed = 408.1920076 code units/h
[Level 5 step 1697] average evolution speed = 408.1882549 code units/h
[Level 6 step 3393] average evolution speed = 408.1835905 code units/h
[Level 7 step 6785] average evolution speed = 408.1802183 code units/h
```
...some output truncated...
```
[Level 7 step 6853] average evolution speed = 407.8479777 code units/h
amrex::Abort::10::NaN in specificAdvance !!!
SIGABRT
See Backtrace.10 file for details
Abort(6) on node 10 (rank 10 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 6) - process 10
Rank 10 aborted with code 6: application called MPI_Abort(MPI_COMM_WORLD, 6) - process 10
```
The BinaryBH example does run a lot faster without puncture tracking though. But maybe that is best left for when we optimize the code. 

**Further updates 21/03/25**:
On the A100s, the BinaryBH examples with puncture tracking will fail at step 52 with an out of memory error:
``` 
amrex::Abort::2::CUDA error 2 in file /lus/eagle/projects/grteclyn/amrex/Src/Base/AMReX_Arena.cpp line 210: out of memory !!!
```
Without puncture tracking, this is another `amrex::Abort::1::NaN in specificAdvance !!!` at step 58.

Examples/BinaryBH/BinaryBHLevel.cpp
mirenradia: So, I think I've resolved Katy's problem by removing the magic number `2` in the template argument for `BHAMR` and replacing it with `BinaryBHLevel::num_punctures`.

Are you happy with this now?

mirenradia: I added a `puncture_tracking.disable_writeout` parameter.

mirenradia:
> **Update 21/03/25**: I just wanted to report that I ran the BinaryBH example on `develop` (same set up as above) and got the same error (but this time at timestep 54) so it seems unrelated:
> 
> ```
> [Level 0 step 54] average evolution speed = 408.2414332 code units/h
> [Level 1 step 107] average evolution speed = 408.2205231 code units/h
> [Level 2 step 213] average evolution speed = 408.2047277 code units/h
> [Level 3 step 425] average evolution speed = 408.1965414 code units/h
> [Level 4 step 849] average evolution speed = 408.1920076 code units/h
> [Level 5 step 1697] average evolution speed = 408.1882549 code units/h
> [Level 6 step 3393] average evolution speed = 408.1835905 code units/h
> [Level 7 step 6785] average evolution speed = 408.1802183 code units/h
> [Level 7 step 6853] average evolution speed = 407.8479777 code units/h
> amrex::Abort::10::NaN in specificAdvance !!!
> SIGABRT
> See Backtrace.10 file for details
> Abort(6) on node 10 (rank 10 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 6) - process 10
> Rank 10 aborted with code 6: application called MPI_Abort(MPI_COMM_WORLD, 6) - process 10
> ```

I think at this point it would be worth opening a new issue for this and moving the conversation there.

> The BinaryBH example does run a lot faster without puncture tracking though. But maybe that is best left for when we optimize the code.

I used the tiny profiler a while ago and it seemed fine. It might be because the puncture tagging increases the size of the finer levels. It would be worth checking with the tiny profiler again to be sure.

Source/BlackHoles/PunctureTracker.impl.hpp
```diff
@@ -0,0 +1,409 @@
...
+                               });
+        }
+        d_level_puncture_coords.copyToHost(h_level_puncture_coords.data(),
+                                           num_puncture_coords);
+
+        // Sum reduce over levels in case there are punctures on multiple levels
+        // on this proc
+        std::transform(m_puncture_coords.begin(), m_puncture_coords.end(),
+                       h_level_puncture_coords.begin(),
+                       m_puncture_coords.begin(), std::plus<amrex::Real>());
```
Pennycook: Could you hoist the allocations out of this loop, and do the summation on the GPU instead?

I'm not sure exactly how the indexing works in AMReX, but I'm imagining something with a structure like:

```c++
// Set up the allocation.
amrex::Gpu::AsyncArray<amrex::ParticleReal> d_level_puncture_coords(num_puncture_coords);
auto *d_level_puncture_coords_ptr = d_level_puncture_coords.data();

// Zero the allocation on the GPU
amrex::ParallelFor(..., [=]() {
  d_level_puncture_coords_ptr[...] = 0;
});

// Sum over all the levels
// NB: If AMReX supports launching a kernel over all the levels, that would be even better
for (int ilevel = 0; ilevel <= m_gr_amr->finestLevel(); ilevel++)
{
  ...
  // Note this is a += now, to do what the std::transform does
  d_level_puncture_coords_ptr[linear_idx(punc_idx, idir)] += p.pos(idir);
  ...
}

// Copy the result back for the MPI communication
d_level_puncture_coords.copyToHost(h_level_puncture_coords.data(), num_puncture_coords);
```

julianakwan:
I did some lightweight profiling of the puncture tracker using AMReX's `Tiny Profiler`. 

I found that if the regridding was turned off on all levels as in `params_profile.txt` the timing of one coarse time step is about the same with and without puncture tracking to within 10%. I checked this with 1 MPI rank to 1 PVC tile and 12 MPI ranks to 12 PVC tiles. I also do not output plotfiles or checkpoints. The only outputs are the locations of punctures if puncture tracking is enabled. 

However, if the regridding is on as in `params.txt`, then the puncture tracking could take as much as 2x the normal runtime. (again, checked with the same configurations). In this test, I have also disabled plotfiles and checkpointing and the only outputs are the punctures. I run the configuration in `params.txt` for two coarse timesteps. 

Here are the outputs of `Tiny Profiler` for these four configurations with and without puncture tracking. I noticed that a lot more device memory is being used with puncture tracking enabled and a key difference between `params.txt` and `params_profile.txt` is that regridding is turned on for every level so there could just be more finer levels being generated with the puncture tracker. 

`Tiny Profiles` with puncture tracking enabled on level 5, configurations in [`params_profile.txt`](https://github.com/GRTLCollaboration/GRTeclyn/blob/enhancement/puncture_tracker2/Examples/BinaryBH/params_profile.txt) and [`params.txt`](https://github.com/GRTLCollaboration/GRTeclyn/blob/enhancement/puncture_tracker2/Examples/BinaryBH/params.txt), except `plot_interval=-1` and `checkpoint_interval=-1`, `max_steps=2`: 
[puncture_tracking_params_profile_pvc_1_mpi_rank_1_gpu.txt](https://github.com/user-attachments/files/19494037/puncture_tracking_params_profile_pvc_1_mpi_rank_1_gpu.txt)
[puncture_tracking_params_profile_pvc_12_mpi_rank_12_gpu.txt](https://github.com/user-attachments/files/19494038/puncture_tracking_params_profile_pvc_12_mpi_rank_12_gpu.txt)
[puncture_tracking_params_pvc_1_mpi_rank_1_gpu.txt](https://github.com/user-attachments/files/19494039/puncture_tracking_params_pvc_1_mpi_rank_1_gpu.txt)
[puncture_tracking_params_pvc_12_mpi_rank_12_gpu.txt](https://github.com/user-attachments/files/19494040/puncture_tracking_params_pvc_12_mpi_rank_12_gpu.txt)

`Tiny Profiles` without puncture tracking, configurations in [`params_profile.txt`](https://github.com/GRTLCollaboration/GRTeclyn/blob/develop/Examples/BinaryBH/params_profile.txt) and [`params.txt`](https://github.com/GRTLCollaboration/GRTeclyn/blob/develop/Examples/BinaryBH/params.txt), except `plot_interval=-1` and `checkpoint_interval=-1`, `max_steps=2`: 
[params_profile_pvc_1_mpi_rank_1_gpu.txt](https://github.com/user-attachments/files/19494086/params_profile_pvc_1_mpi_rank_1_gpu.txt)
[params_profile_pvc_12_mpi_rank_12_gpu.txt](https://github.com/user-attachments/files/19494087/params_profile_pvc_12_mpi_rank_12_gpu.txt)
[params_pvc_1_mpi_rank_1_gpu.txt](https://github.com/user-attachments/files/19494088/params_pvc_1_mpi_rank_1_gpu.txt)
[params_pvc_12_mpi_rank_12_gpu.txt](https://github.com/user-attachments/files/19494089/params_pvc_12_mpi_rank_12_gpu.txt)

mirenradia:
> However, if the regridding is on as in `params.txt`, then the puncture tracking could take as much as 2x the normal runtime. (again, checked with the same configurations).

As discussed on Slack, this is not surprising given the puncture tagger tags more cells for refinement around the punctures. This is necessary in order to properly resolve the black hole horizons and ensure accurate simulations.

Source/BlackHoles/PunctureTracker.impl.hpp
mirenradia: Thanks for the suggestion.

I think I've basically done it by switching the type of the device puncture coords array to `amrex::Gpu::DeviceVector` instead of `amrex::Gpu::AsyncArray`. The constructor of this type initializes all the values to `0.0` on the GPU which saves us having to copy an array of 0s from host to device. I've also moved the sum reduction over levels onto the GPU as you suggested.

I don't think it's possible to launch the kernel over multiple levels because of the way the particle iterators work. Since the number of puncture particles is expected to be small (typically just 2), most of the time only one of the iterations over levels will actually run since the rest will be skipped by
```cpp
        if (this->NumberOfParticlesAtLevel(ilevel) == 0L)
        {
            continue;
        }
```
so I don't think it matters too much.

Source/BlackHoles/PunctureTracker.impl.hpp
mirenradia: I think this should probably be implemented in `SmallDataIO` rather than here so will open a PR to do so.

mirenradia: #109

Examples/BinaryBH/BinaryBHLevel.cpp
mirenradia: I added a `puncture_tracking.writeout_level` parameter to control this.

mirenradia:
I think I've resolved all of your suggestions, @julianakwan including adding particles to the regression test using the AMReX `particle_compare.exe` tool. Please could you re-review?

julianakwan (approved):
Thanks for addressing the feedback on the PR! I have checked that the puncture tracker will: 
-  run on PVCs and A100s 
-  perform a regression test with particle output comparison
-  perform an unit test 
-  work with all example input files in the `BinaryBH` directory
-  output particles in a format that can be visualized with ParaView
-  only write out the puncture locations once for each coarse time step
-  not write out the punctures when `disable_writeout=true`

PR #94, opened 2025-03-01
Matter: Port Matter classes (with unit tests)

### LINKED ISSUE #10: Port Matter classes
It would make sense to also port the ScalarField example at the same time.

julianakwan:
This PR completes my port of the `Matter` classes and closes #10 . It includes:

-  Porting `MatterCCZ4`
-  Porting the `ScalarField` and `DefaultPotential` classes
-  Porting the calculation of `Weyl4_Re` and `Weyl4_Im` for a `Matter` class
-  Porting the calculation of the energy density, the momentum density and the spatial stress energy density for `Matter` classes
-  Porting the calculation of the relaxation of chi (I didn't know it was going to be depreciated!)
-  Unit tests for the above
-  Renaming `NewMatterConstraints` to just `MatterConstraints`.
-  The removal of the `Matter` directory from the list of ignored files in `.lint_ignore`


Tests/ChiRelaxationTest/ChiRelaxationTest.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+// AMReX headers
+#include "AMReX.H"
+#include "AMReX_FArrayBox.H"
+#include "AMReX_MultiFab.H"
+
+#ifdef AMREX_USE_HDF5
+#include "H5Cpp.h"
+#include <AMReX_PlotFileUtilHDF5.H>
+using namespace H5;
```
mirenradia: I think we can remove this since `AMReX_PlotFileUtilHDF5.H` already includes the
HDF5 C header. Since the HDF5 library can be built without C++ support, we
shouldn't unnecessarily include it.

In general, I don't like `using namespace xxx` statements as well since this
breaks namespace clash protection.
```suggestion
#include <AMReX_PlotFileUtilHDF5.H>
```

Tests/ChiRelaxationTest/ChiRelaxationTest.cpp
```diff
@@ -0,0 +1,173 @@
...
+            amrex::Abort("NaN found in output multifab");
+            CHECK(false);
+        }
+        else
+        {
+            CHECK(true);
+        }
+
+#if AMREX_USE_HDF5
+        const H5std_string grteclyn_hdf5_file =
```
mirenradia: With the removal of the `H5Cpp.H` header I suggested above...
```suggestion
        const std::string grteclyn_hdf5_file =
```
This is also the type in the signature of `amrex::WriteSingleLevelPlotfileHDF5`.

Tests/ChiRelaxationTest/ChiRelaxationTest.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+        if (out_mf.contains_nan())
+        {
+            amrex::Abort("NaN found in output multifab");
+            CHECK(false);
+        }
+        else
+        {
+            CHECK(true);
+        }
```
mirenradia: This seems a bit verbose. Why can't we just do the following?
```suggestion
        CHECK(!out_mf.contains_nan());
```

Source/Matter/ChiRelaxation.hpp
mirenradia, line 1: Maybe, for consistency we should rename this class (and file) to
`ChiRelaxationRHS`?

As I have suggested for `MatterCCZ4RHS`, we can also probably remove the
`matter_t` object from the constructor since it's default constructible.

Source/Matter/ChiRelaxation.impl.hpp
```diff
@@ -20,30 +20,33 @@ ChiRelaxation<matter_t>::ChiRelaxation(matter_t a_matter, double dx,
 
 template <class matter_t>
 template <class data_t>
-void ChiRelaxation<matter_t>::compute(Cell<data_t> current_cell) const
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+ChiRelaxation<matter_t>::compute(int i, int j, int k,
+                                 const amrex::Array4<data_t> &rhs,
```
mirenradia: Maybe we should rename this argument to `rhs_arrays` (and probably the next one
to `state_arrays` for consistency). Then the `Vars<data_t>` object in the
function can be `rhs_vars` or `vars_rhs`. I think this is a bit clearer.

Alternatively, I'm happy leaving the renaming of this argument to future coding
convention cleanup and instead just rename the `Vars<data_t>` object.

Source/Matter/EMTensor.hpp
```diff
@@ -19,22 +19,46 @@ template <class matter_t> class EMTensor
...
   public:
     template <class data_t>
-    using Vars = typename MatterCCZ4<matter_t>::template Vars<data_t>;
+    using Vars = typename MatterCCZ4RHS<matter_t>::template Vars<data_t>;
+
+    /// derive record name
+    static inline const std::string name = "EMTensor";
+
+    /// Variable names
+    static inline amrex::Vector<std::string> var_names = {"Rho"};
```
mirenradia: It's a lowercase $\rho$ I believe.
```suggestion
    static inline amrex::Vector<std::string> var_names = {"rho"};
```

Source/Matter/EMTensor.hpp
```diff
@@ -19,22 +19,46 @@ template <class matter_t> class EMTensor
...
+
+    /// 3 components for the momentum density: m_11, m_22, m_33
+    /// 6 components for the stress energy tensor (symmetric): s_11, s_12 etc.
+    static inline const amrex::Vector<std::string> extra_var_names = {
+        "m_11, m_22, m_33, s_11, s_12, s_13, s_22, s_23, s_33"};
```
mirenradia: @KAClough, what do we want to call the momentum density and stress tensor? I
probably would have used uppercase S for the latter but know there are a few
different conventions for the former.

Source/Matter/EMTensor.impl.hpp
mirenradia, line 1: I would like to redesign this class to do the following:
* update it since we now use it in a different way to how it was in GRChombo.
  Having the component numbers/intervals in the constructor doesn't really make
  sense anymore.
* allow the user to decide at compile time (i.e. when creating their example)
  which components they want this to compute (i.e. which of energy density,
  momentum density and stress tensor they want) and then remove any runtime
  branching in the compute class (can change `if`s to `if constexpr`).

I think I might play around myself with this whilst you look at my other comments.

Source/Matter/Make.package
mirenradia, line 1: Let's make this one line per file and in alphabetical order like I've done for
some of the other ones (see `../CCZ4/Make.package`).

Source/Matter/MatterCCZ4RHS.impl.hpp
```diff
@@ -23,15 +23,17 @@ MatterCCZ4RHS<matter_t, gauge_t, deriv_t>::MatterCCZ4RHS(
 
 template <class matter_t, class gauge_t, class deriv_t>
 template <class data_t>
-void MatterCCZ4RHS<matter_t, gauge_t, deriv_t>::compute(
-    Cell<data_t> current_cell) const
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+MatterCCZ4RHS<matter_t, gauge_t, deriv_t>::compute(
+    int i, int j, int k, const amrex::Array4<data_t> &rhs,
```
mirenradia: Cf. my comment for the `ChiRelaxtion` class. We should consistently rename the
`Array4` arguments and the `Vars<data_t>` objects in the function.

Source/Matter/MatterConstraints.impl.hpp
mirenradia, line 1: Like my comment on the `EMTensor` class, I would like to redesign this class similarly.

Source/Matter/MatterWeyl4.impl.hpp
```diff
@@ -12,16 +12,20 @@
...
-    const auto vars = current_cell.template load_vars<Vars>();
-    const auto d1   = m_deriv.template diff1<Vars>(current_cell);
-    const auto d2   = m_deriv.template diff2<Diff2Vars>(current_cell);
+    const auto vars = load_vars<Vars>(state.cellData(i, j, k));
+    const auto d1   = m_deriv.template diff1<Vars>(i, j, k, state);
+    const auto d2   = m_deriv.template diff2<Diff2Vars>(i, j, k, state);
 
     // Get the coordinates
-    const Coordinates<data_t> coords(current_cell, m_dx, m_center);
+    amrex::IntVect cell_coords(AMREX_D_DECL(i, j, k));
```
mirenradia: Let's call this something different to avoid confusion with the `coords` object
below. Alternatively, the `IntVect` can just be constructed in place in the next
line since it's not used again I think.

Source/Matter/MatterWeyl4.impl.hpp
```diff
@@ -12,16 +12,20 @@
 
 template <class matter_t>
 template <class data_t>
-void MatterWeyl4<matter_t>::compute(Cell<data_t> current_cell) const
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+MatterWeyl4<matter_t>::compute(int i, int j, int k,
+                               const amrex::Array4<data_t> &derive,
+                               const amrex::Array4<data_t const> &state) const
```
mirenradia: Let's call both of these arguments `xxx_array`s.

Source/Matter/MatterWeyl4.impl.hpp
```diff
@@ -43,21 +47,20 @@ void MatterWeyl4<matter_t>::compute(Cell<data_t> current_cell) const
...
-                                          const Tensor<2, data_t> &h_UU,
-                                          const chris_t<data_t> &chris) const
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE void MatterWeyl4<matter_t>::add_matter_EB(
+    EBFields_t<data_t> &ebfields, const Vars<data_t> &vars,
+    const Vars<Tensor<1, data_t>> &d1, const Tensor<3, data_t> &epsilon3_LUU,
+    const Tensor<2, data_t> &h_UU, const chris_t<data_t> &chris) const
 {
     // Calculate decomposed energy momentum tensor components
-    const auto emtensor = m_matter.compute_emtensor(vars, d1, h_UU, chris.ULL);
+    const auto emtensor = MatterWeyl4<matter_t>::m_matter.compute_emtensor(
```
mirenradia: Does this code work? The `m_matter` object doesn't seem to be static?

Source/Matter/MatterWeyl4.impl.hpp
```diff
@@ -70,4 +73,55 @@ void MatterWeyl4<matter_t>::add_matter_EB(EBFields_t<data_t> &ebfields,
...
+    const auto &src_arrays = src_mf.const_arrays();
+
+    GRParmParse pp;
+    std::array<double, AMREX_SPACEDIM> center{};
+    int formulation      = 0;
+    amrex::Real G_newton = 0;
+
+    pp.get("extraction_center", center);
+    pp.get("formulation", formulation);
+    pp.get("G_newton", G_newton, 0);
```
mirenradia: As discussed in-person, the final argument is not the default value as for
`GRParmParse::load()`.

Source/Matter/ScalarField.hpp
```diff
@@ -32,12 +32,14 @@
 template <class potential_t = DefaultPotential> class ScalarField
 {
   protected:
-    //! The local copy of the potential
     potential_t my_potential;
+    //! The local copy of the potential
```
mirenradia:
```suggestion
    potential_t m_potential;
```

Source/Matter/ScalarField.hpp
```diff
@@ -32,12 +32,14 @@
...
 
   public:
+
     //!  Constructor of class ScalarField, inputs are the matter parameters.
     ScalarField(const potential_t a_potential) : my_potential(a_potential) {}
```
mirenradia: Let's remove this constructor if we're not using it.
```suggestion
```

Tests/BSSNMatterTest/BSSNMatterTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+                // the initial data doesn't include phi or Pi so do it here:
+                in_array[ibox](i, j, k, c_phi) =
+                    0.34578 + 0.26898 * x + 0.54348 * x * x +
+                    0.33487 * x * y * y * y + 0.79469 * y * z +
+                    0.30515 * z * z + 1.88385 * z * z * z * z;
+                in_array[ibox](i, j, k, c_Pi) =
+                    0.65668 + 0.20188 * x + 0.34348 * x * x +
+                    0.31787 * x * y * y * y + 0.88469 * y * z +
+                    0.10515 * z * z + 1.88385 * z * z * z * z;
```
mirenradia: Since this is used for multiple tests, can you add a Common function for it like
the `random_ccz4_initial_data` one?

Source/Matter/MatterCCZ4RHS.hpp
mirenradia, line 1: If we're going to require our `matter_t` objects to be default constructible, we
can probably just remove it from the constructor for this class.

Tests/BSSNMatterTest/BSSNMatterTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+        if (out_mf.contains_nan())
+        {
+            amrex::Abort("NaN found in output multifab");
+            CHECK(false);
+        }
+        else
+        {
+            CHECK(true);
+        }
```
mirenradia:
```suggestion
        CHECK(!out_mf.contains_nan());
```

Tests/BSSNMatterTest/BSSNMatterTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+// AMReX headers
+#include "AMReX.H"
+#include "AMReX_FArrayBox.H"
+#include "AMReX_MultiFab.H"
+
+#ifdef AMREX_USE_HDF5
+#include "H5Cpp.h"
+#include <AMReX_PlotFileUtilHDF5.H>
+using namespace H5;
```
mirenradia:
```suggestion
#include <AMReX_PlotFileUtilHDF5.H>
```

Tests/BSSNMatterTest/BSSNMatterTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+        {
+            CHECK(true);
+        }
+
+#if AMREX_USE_HDF5
+
+        amrex::Vector<std::string> var_names = ArrayTools::concatenate(
+            StateVariables::names, Constraints::var_names);
+
+        const H5std_string grteclyn_hdf5_file = "BSSNMatterTest/BSSNMatterTest";
```
mirenradia:
```suggestion
        const std::string grteclyn_hdf5_file = "BSSNMatterTest/BSSNMatterTest";
```

Tests/ChiRelaxationTest/ChiRelaxationTest.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+                // the initial data doesn't include phi or Pi so do it here:
+                in_array[ibox](i, j, k, c_phi) =
+                    0.34578 + 0.26898 * x + 0.54348 * x * x +
+                    0.33487 * x * y * y * y + 0.79469 * y * z +
+                    0.30515 * z * z + 1.88385 * z * z * z * z;
+                in_array[ibox](i, j, k, c_Pi) =
+                    0.65668 + 0.20188 * x + 0.34348 * x * x +
+                    0.31787 * x * y * y * y + 0.88469 * y * z +
+                    0.10515 * z * z + 1.88385 * z * z * z * z;
```
mirenradia: See my comment for the BSSNMatterTest

Tests/EMTensorTest/Make.package
```diff
@@ -0,0 +1,3 @@
+GRTECLYN_CEXE_headers += EMTensorTest.hpp
```
mirenradia:
```suggestion
GRTECLYN_CEXE_headers += EMTensorTest.hpp
```

Tests/EMTensorTest/EMTensorTest.cpp
```diff
@@ -0,0 +1,161 @@
...
+#include <AMReX.H>
+#include <AMReX_FArrayBox.H>
+#include <AMReX_FabArray.H>
+#include <AMReX_MFIter.H>
+#include <AMReX_MultiFab.H>
+
+#ifdef AMREX_USE_HDF5
+#include "H5Cpp.h"
+#include <AMReX_PlotFileUtilHDF5.H>
+using namespace H5;
```
mirenradia:
```suggestion
#include <AMReX_PlotFileUtilHDF5.H>
```

Tests/EMTensorTest/EMTensorTest.cpp
```diff
@@ -0,0 +1,161 @@
...
+        int *bcrec  = nullptr;
+        int level   = 0;
+
+        EMTensor<DefaultScalarField>::compute_mf(
+            out_mf, dcomp_rho, num_comps, in_mf, geom, time, bcrec, level);
+
+#if AMREX_USE_HDF5
+        amrex::Vector<std::string> var_names = {"rho"};
+
+        const H5std_string grteclyn_hdf5_file = "EMTensorTest/EMTensorTestOut";
```
mirenradia:
```suggestion
        const std::string grteclyn_hdf5_file = "EMTensorTest/EMTensorTestOut";
```

Tests/EMTensorTest/EMTensorTest.cpp
```diff
@@ -0,0 +1,161 @@
...
+// Doctest includes
+#include "doctest.h"
+#include "doctestCLIArgs.hpp"
+
+#include <iomanip>
+#include <iostream>
+#include <sys/time.h>
+
+// common includes
+#include <InitialData.hpp> //includes StateVariables.hpp
```
mirenradia:
```suggestion
#include "InitialData.hpp" //includes StateVariables.hpp
```

Tests/EMTensorTest/EMTensorTest.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+// test header
+#include "EMTensorTest.hpp"
+
+// GRTeclyn includes
+
```
mirenradia:
```suggestion
```

Tests/EMTensorTest/EMTensorTest.cpp
```diff
@@ -0,0 +1,161 @@
...
+                // Theta is zero for BSSN
+                in_arrays[box_no](i, j, k, c_Theta) = 0.0;
+
+                // the initial data doesn't include phi or Pi so do it here:
+                in_arrays[box_no](i, j, k, c_phi) =
+                    0.21232 * sin(x * 2.1232 * 3.14) * cos(y * 2.5123 * 3.15) *
+                    cos(z * 2.1232 * 3.14);
+                in_arrays[box_no](i, j, k, c_Pi) =
+                    0.4112 * sin(x * 4.123 * 3.14) * cos(y * 2.2312 * 3.15) *
+                    cos(z * 2.5123 * 3.14);
```
mirenradia: This is the same as the `MatterWeyl4Test` so maybe make a common function like `random_ccz4_initial_data()`?

Tests/MatterWeyl4Test/Make.package
```diff
@@ -0,0 +1,3 @@
+GRTECLYN_CEXE_headers += MatterWeyl4Test.hpp
```
mirenradia:
```suggestion
GRTECLYN_CEXE_headers += MatterWeyl4Test.hpp
```

Tests/MatterWeyl4Test/MatterWeyl4Test.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include <AMReX.H>
+#include <AMReX_FArrayBox.H>
+#include <AMReX_FabArray.H>
+#include <AMReX_MFIter.H>
+#include <AMReX_MultiFab.H>
+
+#ifdef AMREX_USE_HDF5
+#include "H5Cpp.h"
+#include <AMReX_PlotFileUtilHDF5.H>
+using namespace H5;
```
mirenradia:
```suggestion
#include <AMReX_PlotFileUtilHDF5.H>
```

Tests/MatterWeyl4Test/MatterWeyl4Test.cpp
```diff
@@ -0,0 +1,170 @@
...
+// Doctest includes
+#include "doctest.h"
+#include "doctestCLIArgs.hpp"
+
+#include <iomanip>
+#include <iostream>
+#include <sys/time.h>
+
+// common includes
+#include <InitialData.hpp> //includes StateVariables.hpp
```
mirenradia:
```suggestion
#include "InitialData.hpp" //includes StateVariables.hpp
```

Tests/MatterWeyl4Test/MatterWeyl4Test.cpp
```diff
@@ -0,0 +1,170 @@
...
+                // Theta is zero for BSSN
+                in_arrays[box_no](i, j, k, c_Theta) = 0.0;
+
+                // the initial data doesn't include phi or Pi so do it here:
+                in_arrays[box_no](i, j, k, c_phi) =
+                    0.21232 * sin(x * 2.1232 * 3.14) * cos(y * 2.5123 * 3.15) *
+                    cos(z * 2.1232 * 3.14);
+                in_arrays[box_no](i, j, k, c_Pi) =
+                    0.4112 * sin(x * 4.123 * 3.14) * cos(y * 2.2312 * 3.15) *
+                    cos(z * 2.5123 * 3.14);
```
mirenradia: See my comment for the EMTensorTest.

Tests/MatterWeyl4Test/MatterWeyl4Test.cpp
```diff
@@ -0,0 +1,170 @@
...
+        pp.queryAdd("formulation", formulation);
+        pp.queryAdd("G_newton", G_Newton);
+
+        MatterWeyl4<DefaultScalarField>::compute_mf(out_mf, dcomp_weyl4,
+                                                    num_comps_weyl4, in_mf,
+                                                    geom, time, bcrec, level);
+
+#if AMREX_USE_HDF5
+        const H5std_string grteclyn_hdf5_file =
+            "MatterWeyl4Test/MatterWeyl4TestOut";
```
mirenradia:
```suggestion
        const std::string grteclyn_hdf5_file =
            "MatterWeyl4Test/MatterWeyl4TestOut";
```

Tests/TestCases.hpp
```diff
@@ -12,19 +12,41 @@
...
 #include "ConstraintsTest.hpp"
 #include "CoordinateTransformationsTest.hpp"
 #include "DerivativeUnitTests.hpp"
+#include "EMTensorTest.hpp"
+#include "MatterWeyl4Test.hpp"
 #include "PositiveChiAndAlphaUnitTest.hpp"
 #include "SphericalHarmonicTest.hpp"
 #include "Weyl4Test.hpp"
 
+TEST_CASE("BSSNMatter"
```
mirenradia, line 30: For consistency with the other test names, let's use spaces.
```suggestion
TEST_CASE("BSSN Matter"
```

Tests/TestCases.hpp
```diff
@@ -12,19 +12,41 @@
...
+)
+{
+    run_bssn_matter_test();
+}
+
 TEST_CASE("CCZ4 Geometry") { run_ccz4_geometry_unit_tests(); }
 
 TEST_CASE("CCZ4 RHS") { run_ccz4_rhs_test(); }
 
+TEST_CASE("ChiRelaxation"
```
mirenradia:
```suggestion
TEST_CASE("Chi Relaxation"
```

Tests/TestCases.hpp
```diff
@@ -41,6 +63,24 @@ TEST_CASE("Coordinate Transformations")
 
 TEST_CASE("Derivative Unit Tests") { run_derivative_unit_tests(); }
 
+TEST_CASE("EMTensor"
```
mirenradia, line 59:
```suggestion
TEST_CASE("EM Tensor"
```

Tests/TestCases.hpp
```diff
@@ -41,6 +63,24 @@ TEST_CASE("Coordinate Transformations")
...
+#ifndef AMREX_USE_HDF5
+          * doctest::skip()
+#endif
+)
+{
+    run_emtensor_test();
+}
+
+TEST_CASE("MatterWeyl4"
```
mirenradia:
```suggestion
TEST_CASE("Matter Weyl4"
```

mirenradia (changes requested):
Thanks for this PR Juliana.

It looks pretty good but I have some comments. I know there are a fair few of them but I think most of them should be easy to resolve and the more complicated ones I will try and look at myself.

For the GRChombo HDF5 files you have generated for the tests, where is the code you used to create them? I couldn't find a relevant branch in the GRChombo repo.

I don't think the `Tests/MatterWeyl4Test/GRChomboEMTensorTest.h5` file is needed/used so could you remove it as well?

Would you also be able to fix the commit history by applying the `fixup`s?

Source/Matter/ChiRelaxation.hpp
KAClough: Let's just get rid of it! It really doesn't work well and shouldn't be used...

Source/Matter/EMTensor.hpp
KAClough: Yes. We have always used Si and Sij but here that naming might be confusing. We could use J instead, so then it would be J1,J2,J3, S11, S12 etc. Maybe let's leave this for the naming tidy up as it would be best to be consistent throughout the code.

Source/Matter/EMTensor.impl.hpp
KAClough: This would be good. I always found the passing of intervals in the constructor a bit ugly and confusing.

Source/Matter/MatterConstraints.impl.hpp
KAClough: Yes please.

Source/Matter/MatterConstraints.hpp
```diff
@@ -24,10 +24,7 @@
    For an example of a matter_t class see ScalarField. \sa Constraints(),
    ScalarField()
 */
-template <class matter_t>
-class [[deprecated("Use new MatterConstraints class in "
-                   "NewMatterConstraints.hpp")]] MatterConstraints
-    : public Constraints
+template <class matter_t> class MatterConstraints : public Constraints
```
KAClough: Maybe it is nicer for this to be called ConstraintsWithMatter, and similarly CCZ4RHSWithMatter, Weyl4WithMatter etc but we can do this in the tidy up if it is controversial. The advantage is that the files don't then all start with Matter, which is a bit annoying for tab complete. I also find it more readable.

Source/Matter/MatterCCZ4RHS.impl.hpp
```diff
@@ -44,16 +46,17 @@ void MatterCCZ4RHS<matter_t, gauge_t, deriv_t>::compute(
...
-    current_cell.store_vars(matter_rhs);
+    store_vars(rhs.cellData(i, j, k), matter_rhs);
 }
 
 // Function to add in EM Tensor matter terms to CCZ4 rhs
 template <class matter_t, class gauge_t, class deriv_t>
 template <class data_t>
-void MatterCCZ4RHS<matter_t, gauge_t, deriv_t>::add_emtensor_rhs(
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+MatterCCZ4RHS<matter_t, gauge_t, deriv_t>::add_emtensor_rhs(
```
KAClough: This contains a bug in the gauge calculation that was fixed in GRChombo here: https://github.com/GRTLCollaboration/GRChombo/pull/261/files

Let's fix it now before we forget!

Source/Matter/MatterWeyl4.impl.hpp
```diff
@@ -70,4 +73,55 @@ void MatterWeyl4<matter_t>::add_matter_EB(EBFields_t<data_t> &ebfields,
...
+                                       amrex::Real /*time*/,
+                                       const int * /*bcrec*/, int /*level*/)
+{
+    const auto &out_arrays = out_mf.arrays();
+    const auto &src_arrays = src_mf.const_arrays();
+
+    GRParmParse pp;
+    std::array<double, AMREX_SPACEDIM> center{};
+    int formulation      = 0;
+    amrex::Real G_newton = 0;
```
KAClough: Should this be G_Newton?

Source/Matter/ScalarField.hpp
KAClough: Lol this is my fault. I think of the m in class members as being "my" and I put it everywhere in the code.

Source/Matter/ChiRelaxation.hpp
mirenradia: Yeah I now remember that you are not a fan of this and wanted to be rid of it!

Source/Matter/EMTensor.hpp
mirenradia: I think this is the only place a string name for these is needed so we might as well get it right now. Do you want to use uppercase J or lowercase j?

Source/Matter/MatterConstraints.hpp
mirenradia: I would be fine with this change so long as it is done consistently. We might as well do it now given this whole PR is about matter.

Source/Matter/MatterWeyl4.impl.hpp
julianakwan: Thanks! I forgot to change it back!

Source/Matter/MatterWeyl4.impl.hpp
julianakwan: Done. But are you happy with the other `get`s? `load` or `query` might make more sense if we want to avoid a runtime error if the values are not present.

Source/Matter/MatterCCZ4RHS.impl.hpp
julianakwan: I added `rhs_gauge_add_matter_terms` to the `IntegratedMovingPunctureGauge` class but I don't think that one has been ported yet, because it still uses the old `Cell` method of getting data.

Tests/TestCases.hpp
julianakwan: I think the other test names should be changed to just the filenames/directory names as well as I keep forgetting what they are called.

mirenradia: I think I would be happy to remove the spaces as it makes it easier to type out if you just want to run a single test case. However, I'll allow this so long as you open a PR to change the other test case names.

For reference, you can list the test-cases by doing
```
./Tests.<build config.ex -dt-ltc
```

Source/Matter/MatterWeyl4.impl.hpp
mirenradia: I don't want to use `GRParmParse::load` since this will print output every time this class is constructed if the default argument is used.

Maybe it makes sense to set the default first in the relevant variable and then use `ParmParse::queryAdd` so the default values are added to the ParmParse table (which may be used by more classes after #1 is done)?

Source/Matter/ChiRelaxation.hpp
julianakwan: I'm going to move `ChiRelaxtion` and it's unit test to a separate branch (something like `enhancement/chi-relaxation`) and then people who want it can check it out separately.

Source/Matter/EMTensor.impl.hpp
mirenradia: @KAClough, I had a go on the `enhancement/matter_class_mr_suggest` branch. Have a look [here](https://github.com/GRTLCollaboration/GRTeclyn/tree/1053194c6dc703281c91f7b367349148637d93fe/Source/Matter).

I've made the choice of which components to store depend on a template parameter which is a scoped enum `EMTensorParts`. I've also reduced the choice of which combinations of components can be stored and assumed that a user will want to do one of

* store just the energy density $\rho$,
* store both the energy density $\rho$ and the momentum density $j^i$
* store all components of the EM tensor

What do you think?

Source/Matter/EMTensor.hpp
mirenradia: We should make sure our var names are consistent with the [`emtensor_t` struct](https://github.com/GRTLCollaboration/GRTeclyn/blob/8c6aaba60bb5951be3b1247bda474a2dba1c5978/Source/CCZ4/CCZ4Geometry.hpp#L15-L21).

Source/Matter/EMTensor.impl.hpp
KAClough: My first thought was, who is Mr Suggest? 🤣 

I am happy with the reduction in combinations, realistically no one wants just j and not rho. 

We can discuss in the meeting too but why is it better to have the comps chosen as a template parameter rather than an argument that is passed? I guess to avoid runtime branching as you say above. I just wonder if more templates is more complicated... but I don't feel very strongly against it.

I have some comments on naming which I can give now or wait until tidy up time. e.g. I would call things EMTensorComponents etc (not parts). I like "justEnergyDensity" though! I would have "allDensities" for consistency as the last option.

Source/Matter/EMTensor.hpp
KAClough: I guess lower case j is the most common beyond S_i (in Alcubierre for example). I kind of like S_i but then I agree that the j removes the ambiguity with S_ij.

Source/Matter/EMTensor.impl.hpp
mirenradia: > We can discuss in the meeting too but why is it better to have the comps chosen as a template parameter rather than an argument that is passed? I guess to avoid runtime branching as you say above. I just wonder if more templates is more complicated... but I don't feel very strongly against it.

Yeah, it avoids having to have runtime branches in compute kernels which are bad for performance, particularly on GPUs (although this is not too much of a concern as a derived quantity). I don't think the additional template argument complicates the code, especially since it's already templated over `matter_t` anyway! If anything, I would argue that it simplifies it since it makes the `set_up` function (which is the only that needs to called in new examples) have the same signature as that of other derived quantities.

> I have some comments on naming which I can give now or wait until tidy up time. e.g. I would call things EMTensorComponents etc (not parts). I like "justEnergyDensity" though! I would have "allDensities" for consistency as the last option.

I initially decided against "components" because I kept thinking of the individual coordinate components (i.e. there should be 16) but if you don't mind it, I'm happy to switch to it. I'll make the change to `allDensities`.

KAClough: Yeah I see what you mean. I also thought about EMTensorDecomposition... but that seems long? I would prefer it to parts though.

I agree that having the same signature for all derived quantities set up is a good idea, so ok, let's use the templating.

Source/Matter/MatterConstraints.impl.hpp
mirenradia: I think we should defer redesigning this class to another PR since we would have to redesign the `Constraints` class similarly.

mirenradia: Ok, I've created a new issue (#98) so will resolve this thread for this PR.

Tests/TestCases.hpp
julianakwan: I have removed the white spaces from the other test cases :)

julianakwan:
Hi all - I think I have addressed everyone's comments now (and the lint checks have finally passed). Could you please review this PR again and double check that you are happy with everything? 

Main changes are:

- `ChiRelaxation` has been moved to a new branch `enhancement/chi_relaxation`
- Names starting with `Matter` have been moved to `...WithMatter` e.g. `MatterCCZ4RHS` is now `CCZ4RHSWithMatter`
- Constructor classes involving the `Matter` object no longer require the `Matter` object to be passed as an argument because the object is now default constructed from the templated type. 
- `Tests/Common/InitialData.hpp` now has a function to initialize test scalar field and conjugate momentum values
- I have applied the bugfix from GRChombo to the `MovingPunctureGauge`. Because it is a matter thing only, it is stored in it's own matter file `MovingPunctureGaugeWithMatter.hpp`

Source/Matter/EMTensor.hpp
```diff
@@ -7,34 +7,64 @@
...
-    EMTensor(const matter_t &a_matter, const double dx, const int a_c_rho = -1,
+    EMTensor(const double dx, const int a_c_rho = -1,
              const Interval a_c_Si  = Interval(),
              const Interval a_c_Sij = Interval());
 
-    template <class data_t> void compute(Cell<data_t> current_cell) const;
+    template <class data_t>
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+    compute(int i, int j, int k, const amrex::Array4<data_t> &out_mf,
+            const amrex::Array4<const data_t> &in_mf) const;
```
mirenradia: Can we change these names to make it consistent with other compute classes
```suggestion
    compute(int i, int j, int k, const amrex::Array4<data_t> &out_arrays,
            const amrex::Array4<const data_t> &state_arrays) const;
```

Source/Matter/ConstraintsWithMatter.hpp
```diff
@@ -38,32 +38,43 @@ template <class matter_t> class MatterConstraints : public Constraints
...
   protected:
     matter_t my_matter; //!< The matter object, e.g. a scalar field
     double m_G_Newton;  //!< Newton's constant, set to one by default.
 };
 
-#include "NewMatterConstraints.impl.hpp"
+#include "ConstraintsWithMatter.impl.hpp"
 
-#endif /* NEWMATTERCONSTRAINTS_HPP_ */
+#endif /* MATTERCONSTRAINTS_HPP_ */
```
mirenradia:
```suggestion
#endif /* CONSTRAINTSWITHMATTER_HPP_ */
```

Source/Matter/CCZ4RHSWithMatter.hpp
```diff
@@ -108,16 +111,16 @@ class MatterCCZ4RHS : public CCZ4RHS<gauge_t, deriv_t>
...
-#include "MatterCCZ4RHS.impl.hpp"
+#include "CCZ4RHSWithMatter.impl.hpp"
 
 // This is here for backwards compatibility though the MatterCCZ4RHS
 // class should be used in future hence mark as deprecated
 template <class matter_t>
-using MatterCCZ4 [[deprecated("Use MatterCCZ4RHS instead")]] =
-    MatterCCZ4RHS<matter_t>;
+using MatterCCZ4 [[deprecated("Use CCZ4RHSWithMatter instead")]] =
+    CCZ4RHSWithMatter<matter_t>;
```
mirenradia: Sorry. I should have noticed this in my previous review but this is a relic from
GRChombo which can be removed.
```suggestion
```

Source/Matter/EMTensor.hpp
```diff
@@ -7,34 +7,64 @@
...
+
+    /// Variable names
+    static inline amrex::Vector<std::string> var_names = {"rho"};
+
+    /// 3 components for the momentum density: Si_11, Si_22, Si_33
+    /// 6 components for the stress energy tensor (symmetric): Sij_11, Sij_12
+    /// etc.
+    static inline const amrex::Vector<std::string> extra_var_names = {
+        "Sij_11, Sij_22, Sij_33, Sij_11, Sij_12, Sij_13, Sij_22, Sij_23, "
+        "Sij_33"};
```
mirenradia: I think we've agreed on the following names.

```suggestion
    static inline const amrex::Vector<std::string> extra_var_names = {
        "j_1, j_2, j_3, S_11, S_12, S_13, S_22, S_23, S_33"};
```

Source/Matter/CCZ4RHSWithMatter.hpp
```diff
@@ -87,19 +87,22 @@ class MatterCCZ4RHS : public CCZ4RHS<gauge_t, deriv_t>
...
+                      int a_formulation = CCZ4RHS<>::USE_CCZ4,
+                      double a_G_Newton = 1.0);
 
     //!  The compute member which calculates the RHS at each point in the box
     //!  \sa matter_rhs_equation()
-    template <class data_t> void compute(Cell<data_t> current_cell) const;
+    template <class data_t>
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+    compute(int i, int j, int k, const amrex::Array4<data_t> &rhs,
+            const amrex::Array4<data_t const> &state) const;
```
mirenradia: Let's make these consistent with the definition.
```suggestion
    compute(int i, int j, int k, const amrex::Array4<data_t> &rhs_arrays,
            const amrex::Array4<data_t const> &state_arrays) const;
```

Source/Matter/ConstraintsWithMatter.hpp
```diff
@@ -38,32 +38,43 @@ template <class matter_t> class MatterConstraints : public Constraints
...
+                          const Interval &a_c_Moms, int a_c_Ham_abs_terms = -1,
+                          const Interval &a_c_Moms_abs_terms = Interval());
 
     //! The compute member which calculates the constraints at each point in the
     //! box
-    template <class data_t> void compute(Cell<data_t> current_cell) const;
+    template <class data_t>
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+    compute(int i, int j, int k, const amrex::Array4<data_t> &cst,
+            const amrex::Array4<data_t const> &state) const;
```
mirenradia:
```suggestion
    compute(int i, int j, int k, const amrex::Array4<data_t> &out_arrays,
            const amrex::Array4<data_t const> &state_arrays) const;
```

Source/Matter/ConstraintsWithMatter.impl.hpp
```diff
@@ -0,0 +1,125 @@
...
+      m_G_Newton(G_Newton)
+{
+}
+
+template <class matter_t>
+template <class data_t>
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+ConstraintsWithMatter<matter_t>::compute(
+    int i, int j, int k, const amrex::Array4<data_t> &cst,
+    const amrex::Array4<data_t const> &state) const
```
mirenradia: Will need to change the variables in the function below too...
```suggestion
    int i, int j, int k, const amrex::Array4<data_t> &out_arrays,
    const amrex::Array4<data_t const> &state_arrays) const
```

Source/Matter/Weyl4WithMatter.hpp
```diff
@@ -0,0 +1,68 @@
...
+          m_G_Newton(a_G_Newton)
+    {
+    }
+
+    //! The compute member which calculates the wave quantities at each point on
+    //! the grid
+    template <class data_t>
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+    compute(int i, int j, int k, const amrex::Array4<data_t> &derive,
+            const amrex::Array4<data_t const> &state) const;
```
mirenradia:
```suggestion
    compute(int i, int j, int k, const amrex::Array4<data_t> &derive_arrays,
            const amrex::Array4<data_t const> &state_arrays) const;
```

Source/Matter/Weyl4WithMatter.impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+    const auto &src_arrays = src_mf.const_arrays();
+
+    GRParmParse pp;
+    std::array<double, AMREX_SPACEDIM> center{};
+    int formulation      = 0;
+    amrex::Real G_Newton = 0;
+
+    pp.get("extraction_center", center);
+    pp.get("formulation", formulation);
+    pp.load("G_newton", G_Newton, 0.0);
```
mirenradia: I think we shouldn't use `load` since in theory it can print out some output if
it uses the default argument (e.g. if the example doesn't add the default to the
`ParmParse` table).
```suggestion
    pp.queryAdd("G_newton", G_Newton);
```

mirenradia (changes requested):
Thanks for making the changes, Juliana. Just a couple more minor changes.

Source/Matter/EMTensor.hpp
julianakwan: Based on this comment: 
> We should make sure our var names are consistent with the [`emtensor_t` struct](https://github.com/GRTLCollaboration/GRTeclyn/blob/8c6aaba60bb5951be3b1247bda474a2dba1c5978/Source/CCZ4/CCZ4Geometry.hpp#L15-L21).

I thought we had decided to make them consistent with the existing names without matter? If you prefer the j's, I can open a pull request to change them both?

mirenradia: Let's just change it to my suggestion in the `emtensor_t` struct in this PR.

julianakwan:
Hi @mirenradia  - I believe I have addressed all your comments now. Can you please take another look at this PR? Thanks!

Source/Matter/EMTensor.hpp
mirenradia: Could you change the `emtensor_t` struct in CCZ4Geometry.hpp to
```cpp
template <class data_t> struct emtensor_t
{
    Tensor<2, data_t> S;  //!< S_ij = T_ij
    Tensor<1, data_t> j;  //!< j_i = T_ia_n^a
    data_t trS;           //!< trS = S^i_i
    data_t rho;           //!< rho = T_ab n^a n^b
};
```
and modify where this struct appears accordingly?

julianakwan: I opened a new PR for this (#117) because `CCZ4Geometry.hpp` wasn't part of this original PR. I will change in this PR once that has been accepted and merged.

Source/Matter/MovingPunctureGaugeWithMatter.hpp
```diff
@@ -0,0 +1,58 @@
...
+ * This class implements a slightly more generic version of the moving puncture
+ * gauge. In particular it uses a Bona-Masso slicing condition of the form
+ * f(lapse) = -c*lapse^(p-2)
+ * and a Gamma-driver shift condition
+ **/
+
+/// This class adds the matter terms to the RHS of the gauge equation
+/// for the moving puncture gauge
+
+class MovingPunctureGaugeWithMatter : public MovingPunctureGauge
```
KAClough, line 26: In the original code, we just added this function into the MovingPunctureGauge class, and didn't inherit to a matter version... I can see it is a bit inconsistent with the structure of the CCZ4/CCZ4Matter classes, but it means fewer files and I don't think it should cause issues, or am I missing something?

Source/Matter/ScalarField.impl.hpp
```diff
@@ -107,7 +109,8 @@ template <class potential_t>
 template <class data_t, template <typename> class vars_t,
           template <typename> class diff2_vars_t,
           template <typename> class rhs_vars_t>
-void ScalarField<potential_t>::matter_rhs_excl_potential(
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+ScalarField<potential_t>::matter_rhs_excl_potential(
```
KAClough: I know that I did this - separating out the rhs and emtensor with and without potential - but now I can't remember why! I think it was to enable doing complex fields, but now we do that without any inheritance from the real fields, which is tidier. This just adds another long function definition that doesn't do much in the body of the function and that makes the code messier. Maybe we should just combine them?

KAClough (commented):
I am happy with this PR, I made two minor suggestions for you to consider @julianakwan but I don't feel strongly about them, so whatever you think! Let's get this merged and do the tidy up!

Source/Matter/MovingPunctureGaugeWithMatter.hpp
julianakwan: I did this because I thought it was tidier to have everything to do with matter in the `Source/Matter` directory. It doesn't cause any issues otherwise, it was just an issue of consistency so non-matter people can freely ignore the matter stuff if they choose to :)

mirenradia: I'm with Juliana here. Let's keep the consistency between non-matter and matter stuff.

Source/Matter/ScalarField.impl.hpp
mirenradia: I don't mind too much either way. This function is force-inlined so it shouldn't affect performance too much other than we might be able to simplify the expressions a little if the functions are combined.

julianakwan: I have smushed the two together now! I also combined the calculation of `emtensor` with and without potential.

KAClough: Great, yeah I think this is better.

julianakwan:
I have incorporated the changes to the `EMTensor` class from `enhancements/matter_class_mr_suggest` with some changes - I would prefer to call it `EMTensorOptions` rather than `EMTensorParts` as I think this is clearer. (We could also have `EMTensorComputeOptions` but this is a bit long). I didn't do the same for the `ConstraintsWithMatter` class because the `Constraints` parent class needs to be changed as well (which should be done in another PR).

mirenradia:
> I have incorporated the changes to the `EMTensor` class from `enhancements/matter_class_mr_suggest` with some changes - I would prefer to call it `EMTensorOptions` rather than `EMTensorParts` as I think this is clearer. (We could also have `EMTensorComputeOptions` but this is a bit long). I didn't do the same for the `ConstraintsWithMatter` class because the `Constraints` parent class needs to be changed as well (which should be done in another PR).

I'm fine with this. I think @KAClough suggested `EMTensorComponents`. I think she also wanted to change the last option from `all` to `allDensities`. I'll let her respond but I'm otherwise happy to approve this PR.

mirenradia (approved):

KAClough:
I'm happy with EMTensorOptions - I didn't like Parts either. Options is nice as it shows you have options to pick! I would like to make it allDensities, thanks @mirenradia for remembering this. But we can also do that in the tidy up, it is not crucial. Let's merge this!

PR #97, opened 2025-03-17
CI: Add headers to linter compilation db

mirenradia:
This should make the linting of header files a bit more robust.

I've also added an extra step for the compilation database to be uploaded as artifact so that it can be inspected if necessary.


julianakwan (approved):
Thanks for adding these! I can confirm that the lint errors related to undefined references to `phi` and `Pi` are gone now from the `enhancement/matter_class_fix` branch.

PR #108, opened 2025-03-20
CI: Remove whitespace between test case names

julianakwan:
I have removed the white spaces between the names of the `doc-db` test cases so that they match the file names used for the tests. This is will also make it more convenient to type the names as we can now leave out the quotation marks. 

In the case of `Positive Chi and Alpha`, I made all the words capitalized for readability. 

This is linked to the PR on the `enhancement/matter_class_fix`.


mirenradia (approved):
LGTM

PR #109, opened 2025-04-01
SmallDataIO: Rename old files instead of overwriting

mirenradia:
This replicates the behaviour of AMReX when it comes to checkpoints and plotfiles.


julianakwan (approved):
Thanks for taking the time to add this feature! I think adding this ability to SmallDataIO will have other uses as well beyond the puncture tracker.

PR #117, opened 2025-04-16
Rename components of emtensor_t for consistency

julianakwan:
As PR #94 changes the naming of the `EMTensor` components (used by AMReX in its list of derived variables), I have made the corresponding changes to the `emtensor_t` type so they match for consistency. 

The spatial component of the stress energy tensor `Sij` is now referred to as `S` and the momentum vector `Si` is now `j`.

I will make the necessary changes in the `EMTensor` class as a part of PR#94 as well.


mirenradia (approved):
LGTM

PR #118, opened 2025-04-23
Port ScalarField example from GRChombo

julianakwan:
This is a port of the scalar field example from GRChombo. 

It uses the ported (and renamed) `matter` classes from GRTeclyn and refactored derived variables. Also ported are the initial conditions within `InitialData/ScalarField` and a `Potential` class with mass (`DefaultPotential` class was included with the `matter` classes).

**About the regression test**
I have also updated the regression test to include a separate job for the scalar field example. The workflow is a bit more complicated now, so I want to record how it works here for future reference (because I will definitely forget). 

There are now three `.yml` files for the regression test:
- `regression_test.yml`
- `regression-tests-setup/action.yml`
- `regression-tests-common.yml`

The main workflow is contained in `regression_test.yml`, which has three jobs: building `fcompare`, running the `BinaryBH` regression test and running the `ScalarField` regression test.  The job the builds `fcompare` runs first and uploads the executable as an artifact. The other two jobs are dependent jobs. 

There is a new directory called `regression-tests-setup` containing a single `.yaml` file which _must_ be named `action.yml`. Because the name itself tells you very little, I have placed `action.yml` in it's own directory. `action.yml` is a [composite action](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action) which contains common steps to both jobs such as checking out AMReX and GRTeclyn and installing dependencies and the compilers.  

`regression-tests-common.yml` is a [reusable workflow](https://docs.github.com/en/actions/sharing-automations/reusing-workflows) that does the actual regression test by calling `regression-tests-setup/action.yml` and then building the example, running the example, downloading `fcompare` as an artifact, applying `fcompare` to the results and finally restarting from checkpoint.

_Why have such a complicated process?_ There are only three ways that I know of to reduce duplication between jobs: caching the results/creating artifacts, reusable workflows and composite actions. Caching everything or uploading everything as an artifact would have taken a lot of storage. Any job that uses a reusable workflow cannot have subsequent steps, i.e. a job callling `uses:` cannot also call `steps:` (see [here](https://docs.github.com/en/actions/sharing-automations/reusing-workflows)). Also, environments cannot be saved from a reusable workflow. For these reasons, I had to split things up so that each of these pieces would work within the constraints imposed by GitHub actions and also reduce the amount of duplication. I suppose everything could have gone into the composite action, but then both jobs would have had to build `fcompare`.  

**Scalar Bubble comparison**
In the scalar bubble case (single Gaussian peak in the center of the simulation volume, width and amplitude specified in `params_test.txt`), I did a comparison between the GRTeclyn and GRChombo to check for correctness. 

The GRChombo version is in the branch `test/compare_scalar_field`

This is a side by side comparison of the two coarse grids, cut through the centre to expose the blob structure (left is GRChombo, right is GRTeclyn):
Screen Shot 2025-04-23 at 4 55 24 PM

This is the fractional difference, calculated as (phi_GRChombo-phi_GRTeclyn)/phi_GRChombo. The full simulation volume is on the left and the normalized distribution of differences is shown in the right histogram. This means that 80% of cell values have less than 0.5% absolute difference between the two codes.
Screen Shot 2025-04-23 at 5 21 30 PM

To do:
- [ ] Rebase onto merged `matter` classes once PR has been accepted
- [ ] Update the regression test for particles in the case of the Binary BH example
- [ ] Replace scalar bubble with oscillaton from [here](https://github.com/KAClough/GRChombo_public/tree/main/Examples/Oscillaton)


mirenradia:
Can we close this PR since #207 has been merged?

julianakwan:
> Can we close this PR since #207 has been merged?

Yes, I am happy to close this.

PR #120, opened 2025-05-16
Fix MPI communicator for serial builds 

julianakwan:
After AMRex [PR#4463](https://github.com/AMReX-Codes/amrex/pull/4463), `MPI_COMM_WORLD` is no longer defined for serial builds. This is breaking our CI because we test for both `USE_MPI=TRUE` and `USE_MPI=FALSE`. 

I have removed the MPI communicator from calls to `amrex::Initialize` since we never define our own MPI communicator and it should default to `MPI_COMM_WORLD` in this case. 

The only exception is `SetupFunctions.hpp` in which I have defined a pre-processor guard for serial builds so that `MPI_COMM_WORLD` is recognized as `amrex::MPI_COMM_WORLD` (as per Weiqun's suggestion). This is because we were passing the other argument to `ParmParse` to avoid using the managed memory Arena. This should be the default behaviour anyway, but I think someone should look into it before I remove it 😄

### DIFF
--- a/Source/GRTeclynCore/SetupFunctions.hpp
+++ b/Source/GRTeclynCore/SetupFunctions.hpp
@@ -44,7 +44,7 @@ void mainSetup(int argc, char *argv[])
     // NOLINTEND(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
     bool use_parm_parse = true;
     // NOLINTNEXTLINE(bugprone-casting-through-void)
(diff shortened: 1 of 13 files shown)


mirenradia (changes requested):
Since `amrex.the_arena_is_managed = false` is now the default in AMReX (since AMReX-Codes/amrex#3438 in Jul 2023), I think we can safely remove the lambda that we pass to `amrex::Initialize` in `SetupFunctions.hpp` and just pass the `argc` and `argv` arguments.

mirenradia (approved):
LGTM!

PR #121, opened 2025-05-29
Add Klein Gordon example

julianakwan:
**Overview**
This PR adds the Klein Gordon example with 3 modes of operation: 
- "Wave" for the wave equation: 
$$\frac{\partial^2 \phi}{\partial t^2} = \nabla^2 \phi$$
- "SineGordon1D" for the 1D Sine Gordon breather solution to the equation:
$$\frac{\partial^2 \phi}{\partial t^2} = \frac{\partial^2 \phi}{\partial x^2} - \sin \phi$$
- "SineGordon3D" for the 3D Sine Gordon pseudo-breather solution to the equation:
$$\frac{\partial^2 \phi}{\partial t^2} = \nabla^2 \phi - \sin \phi$$
 
The latter two examples are stable (only approximately for the 3D case but I have never run it for so long that it collapses), while the wave equation is a spherical wave propagating from the center of the box. 

Note that the CCZ4 variables are not propagated in this example, only $\phi$ (the solution) and $\Pi$ (the first derivative). 

**Wave model**
Select this by entering `model = wave` in the input parameter file. 
You will also need to enter `k_r` as in $\phi (\bf{r},t) = \exp\[i(\bf{k_r \cdot r} - \omega t)\]$ for the traveling wave solution $\phi(\bf{r},t)$ and you have the option of entering a value for the `scalar_mass` which will result in a $m^2 \phi^2$ potential being added to the equation of motion. `scalar_mass` will default to 0 if nothing is chosen. 

**Sine-Gordon model**
Select this by entering either `model=SineGordon1D` or `model=SineGordon3D`. You will need to choose a value for `alpha` (must be between 0 and 1) and the `initial_time` which can't be zero for the 3D case. (NB: I don't know what this means physically, but mathematically the solution just becomes uniform throughout the entire volume). They should be stationary breather solutions, in that they won't propagate across the box but will stay pulsating at the center of the simulation volume. 

Each of these models is represented as a separate object and the `KleinGordonRHS` class is templated over this object for easy access to the potential and variables associated with that model. There are also `calc_analytic_mf_3d` and `calc_analytic_mf_1d` functions templated over the `model_t` type that wrap over the `ParallelFor` construct so the user can set the initial conditions and derived variables easily. 

**Derived variables**
The analytic solution is calculated as a derived variable and has been added to `derive_lst` and is labelled as `phi_analytic` and `Pi_analytic`. It will be automatically outputted with the plotfiles if `amr.derive_plot_vars = ALL` or `amr.derive_plot_vars = analytic_soln`. Note that the two solutions will diverge near the edges of the simulation because the analytic solution hasn't accounted for the effect of the boundary conditions.

I have also updated the example to calculate the energy density as it shows how to access the state variables in the `MultiFab` and the value of the potential from the `model` class. This is labelled as `rho` in the list of derived variables. 

**To Do**
- [x] Add regression test
- [x] Add another example involving the calculation of the energy density.  
- [x]  Refactor `Potential` class - the calculation for the $m^2 \phi^2$ potential isn't called in the calculation of the RHS


Examples/KleinGordon/inputs
```diff
@@ -0,0 +1,62 @@
...
+plot_interval = 1           # (hold over from GRChombo)
+amr.plot_int = 1            # (AMReX variable)
+checkpoint_interval = 0   
+
+amr.checkpoint_files_output = 0  # write checkpoint files or not (overrides all other checkpoint options)
+amr.check_int = 100
+
+amr.plot_vars = ALL
+amr.derive_plot_vars = ALL   #also write out derived parameters
+amr.plot_file = /home/dc-kwan1/rds/rds-dirac-dp002/dc-kwan1/GRTeclyn/KleinGordon/test/plt_
```
Pennycook: I had to make this into a relative path to run on my machine:

```suggestion
amr.plot_file = test/plt_
```

julianakwan: Thanks for pointing this out! I've removed the absolute path.

Examples/KleinGordon/DerivedVariables.hpp
mirenradia, line 1: Could you rename this file and function to something like `AnalyticSolutionDerivedVariable.hpp` and `calc_analytic_solution_mf` respectively or something similar? I think it should be clear that this is for calculating the analytic solution.

.github/workflows/lint.yml
mirenradia, line 1: This file has a few issues

1. We don't need to generate `AMReX_Config.H` twice.
2. We don't need to run `compiledb` twice for the tests directory. Maybe move this to a separate step.
3. In the "Generate config and compilation database for BinaryBH example" step, the `compdb` command near the end will add the header files to the compilation database but this needs to be run after all `compiledb` steps have been run. Can you move these last two lines from this step to a new step after all `compiledb` commands?

Examples/KleinGordon/InitialConditions.hpp
mirenradia, line 1: I don't think this really needs to be a class. The only member variable is `m_initial` and since its meaning differs between the member functions, it makes it a bit confusing. Maybe these can just be functions (with an extra argument for `k_r` or `alpha` as appropriate) inside of a namespace?

Can we include either `KleinGordon` or `KG` in the file and namespace name?

Examples/KleinGordon/KleinGordon.hpp
mirenradia, line 1: I don't think it's necessary to create a derived class from `GRAMR` just to add this printout to the destructor. Can you move the printing out of this info to `runGRTeclyn()` in `Main_KleinGordon.cpp` and remove this class?

Examples/KleinGordon/README.md
```diff
@@ -0,0 +1,26 @@
+This solves the Klein-Gordon equation $\frac{\partial^2 \phi}{\partial t^2} = \nabla \phi + V(\phi)$, based on the wave example in the [AMReX guided tutorials](https://github.com/AMReX-Codes/amrex-tutorials).
```
mirenradia:
```suggestion
This solves the Klein-Gordon equation 
$$\frac{\partial^2 \phi}{\partial t^2} = \nabla \phi + V(\phi),$$
based on the wave example in the [AMReX guided tutorials](https://github.com/AMReX-Codes/amrex-tutorials). 
```

Examples/KleinGordon/DerivedVariables.cpp
mirenradia, line 1: This function is a bit long. Can you move the body of each `model` branch into separate functions to make it a bit easier to read?

Examples/KleinGordon/KleinGordonLevel.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    const auto problo = geom.ProbLoArray();
+    const auto dx     = geom.CellSizeArray();
+
+    std::array<double, AMREX_SPACEDIM> center{};
+
+    amrex::ParmParse pp;
+    pp.query("center", center);
+
+    MultiFab &S_new  = get_new_data(State_Type);
```
mirenradia: For consistency with the rest of the file (and our naming conventions).
```suggestion
    MultiFab &state_new  = get_new_data(State_Type);
```
Yes I know this is a problem with the BinaryBH example but we might as well get it right here!

Examples/KleinGordon/DerivedVariables.hpp
julianakwan: I changed the name of the function, but I would like to leave the filenames as they are because ideally I would have another example of a derived variable calculation involving the grid quantities. I think people thought this one was too simple because it didn't involve the AMReX data containers.

julianakwan:
Hi @mirenradia - just letting you know that this branch is finally ready for review 😅 Even though the CI regression test is broken because CSD3 is down, I have manually run the regression test on Tursa: 

```
 variable name            absolute error            relative error
                                        (||A - B||)         (||A - B||/||A||)
 ----------------------------------------------------------------------------
 level = 0
 phi                                1.065814104e-14           8.945691844e-16
 Pi                                 1.076694289e-12           1.798144842e-13
```
I have checked previously that the GitLab regression test completes correctly as there is more stuff in there now with the two examples being built and run.

Examples/KleinGordon/DerivedVariables.hpp
julianakwan: As promised, there is now more stuff now `DerivedVariables.impl.hpp` (previously named `DerivedVariables.cpp` - I have some templated functions in there now). I hope it makes more sense to keep them together in a single file now.

julianakwan:
For completeness, I also manually ran the regression test on Tursa: 

```
            variable name            absolute error            relative error
                                        (||A - B||)         (||A - B||/||A||)
 ----------------------------------------------------------------------------
 level = 0
 phi                                1.065814104e-14           8.945691844e-16
 Pi                                 1.132316463e-12           1.891037249e-13
 phi_analytic                       2.131628207e-14           1.880677921e-15
 Pi_analytic                                      0                         0
 rho                                3.414157845e-12           1.817697778e-13
```

.github/workflows/data/plt00008_BinaryBH_compare/Header
mirenradia, line 1: Rather than adding `_BinaryBH_` in the middle of the filename and naming the KG one similarly, please can you just put them in separate folders?

.github/workflows/regression-tests/action.yml
mirenradia, line 1: As discussed in our 1:1, can you rename this action (i.e. the parent directory of `action.yml`) to `regression-tests-setup`.

.github/workflows/regression-tests/action.yml
```diff
@@ -0,0 +1,49 @@
+name: Common tasks
```
mirenradia: Maybe give this a name specific to regression tests e.g.
```suggestion
name: Regression Tests setup
```

.github/workflows/regression-tests-common.yml
mirenradia, line 1: Maybe this should be called something like `regression-tests-template.yml` or something like that?

.github/workflows/regression-tests-common.yml
```diff
@@ -0,0 +1,77 @@
+name: Reusable workflow
```
mirenradia: Please give this a better name that's related to the filename.

.github/workflows/regression-tests-common.yml
```diff
@@ -0,0 +1,77 @@
...
+      example:            # this is the name of the example
+        required: true
+        type: string
+
+jobs:
+  run_tests:
+    runs-on: ubuntu-24.04
+    env:
+      AMREX_HOME: ${{ github.workspace }}/amrex
+      EXECUTABLE: main3d.gnu.MPI.ex
```
mirenradia, line 18: Can we make this an argument for when we get around to renaming the base names of examples?

.github/workflows/regression-tests.yml
```diff
@@ -41,69 +37,29 @@ jobs:
...
-        --rel_tol ${REL_TOL}
-        ${GITHUB_WORKSPACE}/GRTeclyn/.github/workflows/data/plt00008_compare
-        ${BINARYBH_EXAMPLE_DIR}/plt00008
-        particles
+    - name: Upload fcompare tool
+      uses: actions/upload-artifact@v4
+      with:
+        name: fcompare.gnu.MPI.ex
+        path: ${{ env.AMREX_HOME }}/Tools/Plotfile/fcompare.gnu.MPI.ex
+        compression-level: 0
```
mirenradia, line 45: Did you choose this to speed up this step? I wouldn't have thought compressing this executable would take all that long?

.github/workflows/regression-tests.yml
```diff
@@ -41,69 +37,29 @@ jobs:
...
+      gcc_version: 14
+      example: BinaryBH
+
+  kleingordon:
+    name: KleinGordon
+    needs: build
+    uses: ./.github/workflows/regression-tests-common.yml
+    with:
+      gcc_version: 14
+      example: KleinGordon
```
mirenradia: Since this is basically the same for each example, can you make it a matrix job which depends on an `example` variable?

mirenradia (changes requested):
Thanks for this PR, @julianakwan. It looks good!

I haven't had time to fully review it but I thought I would make a start and share my comments now so you can work on the requested changes. I have looked at all of the files in the `.github/workflows` directory but nothing further. When you re-request my review, I will look at the other files.

Examples/KleinGordon/GNUmakefile
```diff
@@ -0,0 +1,41 @@
...
+include $(src_pack)
+CEXE_sources += $(GRTECLYN_CEXE_sources)
+CEXE_headers += $(GRTECLYN_CEXE_headers)
+VPATH_LOCATIONS += $(src_dirs)
+INCLUDE_LOCATIONS += $(realpath .) $(src_dirs)
+
+include $(GRTECLYN_HOME)/Tools/GNUMake/Make.rules
+
+# Use a common temporary build directory for all GRTeclyn applications
+TMP_BUILD_DIR ?= $(GRTECLYN_HOME)/tmp_build_dir
```
mirenradia: I got rid of this in #82 in order fix building multiple examples at the same time. Let's remove it here too.

Examples/KleinGordon/GNUmakefile
```diff
@@ -0,0 +1,41 @@
...
+include $(GRTECLYN_HOME)/Tools/GNUMake/Make.rules
+
+GRTECLYN_SOURCE = $(GRTECLYN_HOME)/Source
+
+src_dirs := $(GRTECLYN_SOURCE)/AMRInterpolator \
+            $(GRTECLYN_SOURCE)/Grids \
+            $(GRTECLYN_SOURCE)/GRTeclynCore \
+            $(GRTECLYN_SOURCE)/IO \
+            $(GRTECLYN_SOURCE)/Maths \
+            $(GRTECLYN_SOURCE)/simd \
```
mirenradia: Now that #127 is merged, we need to remove this line.

Examples/KleinGordon/KleinGordonLevel.cpp
```diff
@@ -0,0 +1,186 @@
+#include "KleinGordonLevel.hpp"
+#include "FixedGridsTagger.hpp"
+#include "FourthOrderDerivatives.hpp"
+#include "KleinGordonRHS.hpp"
+#include <numeric>
+
+using namespace amrex;
```
mirenradia: Let's not do this in order to avoid polluting the global namespace.
```suggestion
```

Examples/KleinGordon/KleinGordonLevel.cpp
```diff
@@ -0,0 +1,186 @@
...
+    amrex::ParmParse pp;
+    std::string model{};
+    pp.query("model", model);
+
+    const int ncomp_rho{1}; // only one component associated with energy density
+
+    if (model == "Wave")
+    {
+        derive_lst.add(
+            "rho", amrex::IndexType::TheCellType(), 1,
```
mirenradia: Might as well use the variable defined above rather than a magic number.
```suggestion
            "rho", amrex::IndexType::TheCellType(), ncomp_rho,
```

Examples/KleinGordon/KleinGordonLevel.cpp
```diff
@@ -0,0 +1,186 @@
...
+    const amrex::Vector<std::string> comp_names = {"phi_analytic",
+                                                   "Pi_analytic"};
+
+    int ncomp_analytic{
+        static_cast<int>(comp_names.size())}; // how many derived variables
+
+    derive_lst.add(
+        comp_type, amrex::IndexType::TheCellType(), ncomp_analytic, comp_names,
+        calc_analytic_solution, [=](const amrex::Box &box)
+        { return amrex::grow(box, simParams().num_ghosts); },
```
mirenradia: I don't think you need ghosts to calculate the analytic solution.

Examples/KleinGordon/KleinGordonLevel.cpp
```diff
@@ -0,0 +1,186 @@
...
+    pp.query("model", model);
+
+    const int ncomp_rho{1}; // only one component associated with energy density
+
+    if (model == "Wave")
+    {
+        derive_lst.add(
+            "rho", amrex::IndexType::TheCellType(), 1,
+            calc_energy_density<Wave>, [=](const amrex::Box &box)
+            { return amrex::grow(box, simParams().num_ghosts); },
```
mirenradia: I think you just need 2 ghosts rather than whatever the `num_ghosts` parameter
is set to?

Examples/KleinGordon/KleinGordonLevel.cpp
```diff
@@ -0,0 +1,186 @@
...
+            &amrex::cell_quartic_interp);
+    }
+
+    if (model.find("SineGordon") == 0)
+    {
+        derive_lst.add(
+            "rho", amrex::IndexType::TheCellType(), 1,
```
mirenradia:
```suggestion
            "rho", amrex::IndexType::TheCellType(), ncomp_rho,
```

Examples/KleinGordon/KleinGordonLevel.cpp
```diff
@@ -0,0 +1,186 @@
...
+            calc_energy_density<SineGordon>, [=](const amrex::Box &box)
+            { return amrex::grow(box, simParams().num_ghosts); },
```
mirenradia: Hardcode the number of ghosts (as above).

Examples/KleinGordon/KleinGordonLevel.cpp
```diff
@@ -0,0 +1,186 @@
...
+            &amrex::cell_quartic_interp);
+    }
+
+    derive_lst.addComponent("rho", desc_lst, State_Type, 0, 2);
```
mirenradia:
```suggestion
    derive_lst.addComponent("rho", desc_lst, State_Type, 0, NUM_VARS);
```

Examples/KleinGordon/KleinGordonLevel.cpp
```diff
@@ -0,0 +1,186 @@
...
+    BL_PROFILE("KleinGordonLevel::initData()");
+
+    std::array<double, AMREX_SPACEDIM> center{};
+    std::string model{};
+    amrex::Real initial_time{0.0};
+
+    amrex::ParmParse pp;
+    pp.query("center", center);
+    pp.query("model", model);
+    pp.query("initial_time", initial_time);
```
mirenradia: Can you scope the latter two parameters e.g. `model` -> `klein_gordon.model`? We
can then use scoped `ParmParse`s in some places (though maybe not here if you
need `center`).

Examples/KleinGordon/KleinGordonLevel.cpp
```diff
@@ -0,0 +1,186 @@
...
+    }
+    else
+    {
+        calc_analytic_mf_3d<SineGordon>(state_new, dcomp, geom, current_time);
+    }
+}
+void KleinGordonLevel::specificAdvance()
+{
+    // Usually constraints are enforced here, but we don't have any...
+}
```
mirenradia: I think we can just remove this function (and the corresponding declaration in
the header).
```suggestion
```

Examples/KleinGordon/KleinGordonLevel.cpp
```diff
@@ -0,0 +1,186 @@
...
+    auto const &soln_arrs = a_soln.const_arrays();
+    auto const &rhs_arrs  = a_rhs.arrays();
+
+    amrex::ParmParse pp;
+
+    amrex::Real initial_time{0.0};
+    std::string model{};
+
+    pp.query("model", model);
+    pp.query("initial_time", initial_time);
```
mirenradia: Is this used?

Examples/KleinGordon/KleinGordonLevel.cpp
```diff
@@ -0,0 +1,186 @@
...
+            simParams().sigma, dx, my_model);
+
+        amrex::ParallelFor(
+            a_soln,
+            [=] AMREX_GPU_DEVICE(int box_no, int i, int j, int k) noexcept
+            {
+                klein_gordon_rhs.compute(i, j, k, soln_arrs[box_no],
+                                         rhs_arrs[box_no]);
+            });
+    }
```
mirenradia, line 123: It would be good to exploit C++17 class template argument deduction to get rid
of the duplicated code in these conditional blocks. I think you might need to define some
kind of helper function like the following though:

```cpp
auto get_model = [](const std::string& model) {
    if (model == "Wave") return Wave{};
    else return SineGordon{};
};

auto my_model = get_model(model);
KleinGordonRHS klein_gordon_rhs(simParams().sigma, dx, my_model);

amrex::ParallelFor(
    a_soln,
    [=] AMREX_GPU_DEVICE(int box_no, int i, int j, int k) noexcept
    {
        klein_gordon_rhs.compute(i, j, k, soln_arrs[box_no], rhs_arrs[box_no]);
    });
```
If the helper can be used elsewhere, I would suggest making it a proper function
rather than just a Lambda.

Examples/KleinGordon/DerivedVariables.impl.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+    auto const &arrs = mf_out.arrays();
+
+    amrex::ParallelFor(
+        mf_out, mf_out.nGrowVect(),
+        [=] AMREX_GPU_DEVICE(int box_no, int i, int j, int k) noexcept
+        {
+            amrex::Real x = problo[0] + (i + 0.5) * dx[0] - center[0];
+            amrex::Real y = problo[1] + (j + 0.5) * dx[1] - center[1];
+            amrex::Real z = problo[2] + (k + 0.5) * dx[2] - center[2];
```
mirenradia: Let's use `Coordinates` for this.

Examples/KleinGordon/DerivedVariables.impl.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+    model_t model;
+
+    auto const &arrs = mf_out.arrays();
+
+    amrex::ParallelFor(
+        mf_out, mf_out.nGrowVect(),
+        [=] AMREX_GPU_DEVICE(int box_no, int i, int j, int k) noexcept
+        {
+            amrex::Real x = problo[0] + (i + 0.5) * dx[0] - center[0];
```
mirenradia: Let's use `Coordinates` for this.

Examples/KleinGordon/KleinGordonLevel.cpp
```diff
@@ -0,0 +1,186 @@
...
+    const auto &state_arrs = state_new.arrays();
+
+    const amrex::Real dx         = Geom().CellSize(0);
+    const int current_level      = Level();
+    const amrex::Real box_length = Geom().ProbLength(0);
+    std::array<double, AMREX_SPACEDIM> center{AMREX_D_DECL(0., 0., 0.)};
+    GRParmParse pp;
+    pp.query("center", center);
+
+    FixedGridsTagger my_tagging_criterion{dx, current_level, box_length,
```
mirenradia, line 173: This choice of tagging criterion seems a bit odd for none spherically symmetric configurations (e.g. the SineGordon ones).

I guess this code is more of a simple example to show how things work in GRTeclyn rather than something to do accurate simulations with but it might be worth at least adding a comment here to that effect.

Examples/KleinGordon/KleinGordonLevel.hpp
```diff
@@ -0,0 +1,47 @@
...
+    //! Initialize data at problem start-up.
+    void initData() override;
+
+    //! Advance this level for one step
+
+    void specificEvalRHS(amrex::MultiFab &a_soln, amrex::MultiFab &a_rhs,
+                         const double a_time) override;
+
+    void specificAdvance() override;
+
```
mirenradia:
```suggestion
```

Examples/KleinGordon/KleinGordonRHS.hpp
```diff
@@ -0,0 +1,69 @@
...
+    template <class data_t> struct Diff2Vars
+    {
+        data_t phi;
+
+        template <typename mapping_function_t>
+        AMREX_GPU_DEVICE void enum_mapping(mapping_function_t mapping_function)
+        {
+            VarsTools::define_enum_mapping(mapping_function, c_phi, phi);
+        }
+    };
```
mirenradia: Maybe we should get rid of these now since Katy is getting rid of the rest of
them in #137.

Examples/KleinGordon/KleinGordonRHS.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+    template <class data_t>
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+    compute(int i, int j, int k, const amrex::Array4<data_t const> &input,
+            const amrex::Array4<data_t> &output) const;
+
+  protected:
```
mirenradia: Nobody is going to derive from this class...
```suggestion
  private:
```

Examples/KleinGordon/Main_KleinGordon.cpp
```diff
@@ -0,0 +1,87 @@
+/* GRChombo
+ * Copyright 2012 The GRChombo collaboration.
+ * Please refer to LICENSE in GRChombo's root directory.
+ */
+
+// Our includes
+#include "DefaultLevelFactory.hpp"
+#include "GRParmParse.hpp"
+#include "MultiLevelTask.hpp"
```
mirenradia:
```suggestion
```

Examples/KleinGordon/Main_KleinGordon.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include "SetupFunctions.hpp"
+#include "SimulationParameters.hpp"
+
+// Problem specific includes:
+#include "KleinGordonLevel.hpp"
+
+// System includes
+#include <chrono>
+#include <iostream>
+
```
mirenradia: I don't think we use anything directly from these headers here.
```suggestion
```

Examples/KleinGordon/README.md
```diff
@@ -0,0 +1,18 @@
+This solves the Klein-Gordon equation
+$$\frac{\partial^2 \phi}{\partial t^2} = \nabla \phi + V(\phi)$$,
+based on the wave example in the [AMReX guided tutorials](https://github.com/AMReX-Codes/amrex-tutorials).
```
mirenradia:
```suggestion
based on the wave example in the [AMReX guided tutorials](https://github.com/AMReX-Codes/amrex-tutorials).
```

Examples/KleinGordon/README.md
```diff
@@ -0,0 +1,18 @@
+This solves the Klein-Gordon equation
+$$\frac{\partial^2 \phi}{\partial t^2} = \nabla \phi + V(\phi)$$,
```
mirenradia: An effort to make the LaTeX displayed equation work:
````suggestion

```math
\frac{\partial^2 \phi}{\partial t^2} = \nabla \phi + V(\phi),
```

````

File: Examples/KleinGordon/README.md [hunk 32]
```diff
@@ -0,0 +1,18 @@
...
+based on the wave example in the [AMReX guided tutorials](https://github.com/AMReX-Codes/amrex-tutorials). 
+
+Choose from three possible models in the input parameter file:
+* `model=wave` simulates the wave equation with $m^2 \phi^2$ potential
+* `model=SineGordon1D` simulates the 1D breather solution of the Sine-Gordon equation
+* `model=SineGordon3D` simulates the 3D pseudo-breather solution of the Sine-Gordon equation
```
mirenradia: For completeness, can you write out the potential in the `Sine-Gordon` cases and
specify all model initial/analytic solutions? Please also prepend $V(\phi) = $
to the wave potential.

Examples/KleinGordon/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Other settings in the parameter file:
+* alpha - (only for SineGordon models) controls the frequency of the breather mode
+* initial_time - (only for SineGordon models) intial starting time of the SineGordon breather
+* scalar_mass - (only for Wave model) mass of the scalar field in the potential
```
mirenradia: I guess this will be easier to explain when you write out the initial/analytic solution.
```suggestion
* `alpha` - (only for SineGordon models) controls the frequency of the breather mode
* `initial_time` - (only for SineGordon models) intial starting time of the SineGordon breather
* `scalar_mass` - (only for Wave model) mass of the scalar field in the potential
```

Examples/KleinGordon/SimulationParameters.hpp
```diff
@@ -0,0 +1,72 @@
...
+        else if (model.find("SineGordon") ==
+                 0) // this is for Sine-Gordon ICs only
+        {
+            // These are parameters specfic to the Sine Gordon example
+            pp.queryAdd("alpha", alpha);
+        }
+        else
+        {
+            amrex::Abort("Model option not recognized");
+        }
```
mirenradia, line 66: Let's put these parameters in a "namespace" e.g. `klein_gordon.model`.

Examples/KleinGordon/SineGordon.hpp
```diff
@@ -0,0 +1,103 @@
...
+// KleinGordon includes
+#include "StateVariables.hpp"
+
+class SineGordon
+{
+  public:
+    amrex::Real m_alpha{0.7};
+    amrex::Real m_t0{0.0};
+
+    // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
```
mirenradia: Maybe this is a relic from when the constructor had arguments?
```suggestion
```

Examples/KleinGordon/SineGordon.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+        // Sine Gordon 3D psuedo-breather solution
+
+        return 4.0 * 4.0 * 4.0 *
+               std::atan(m_alpha * std::sin(beta * (t + m_t0)) / beta /
+                         std::cosh(m_alpha * x)) *
+               std::atan(m_alpha * std::sin(beta * (t + m_t0)) / beta /
+                         std::cosh(m_alpha * y)) *
+               std::atan(m_alpha * std::sin(beta * (t + m_t0)) / beta /
+                         std::cosh(m_alpha * z));
```
mirenradia: You might as well precompute `m_alpha * std::sin(beta * (t + m_t0)) / beta`.

Examples/KleinGordon/SineGordon.hpp
```diff
@@ -0,0 +1,103 @@
...
+    // readability-convert-member-functions-to-static)
+    [[nodiscard]] AMREX_GPU_DEVICE AMREX_FORCE_INLINE amrex::Real
+    calculate_derivative(const amrex::Real x, const amrex::Real y,
+                         const amrex::Real z, const amrex::Real t) const
+    // NOLINTEND(bugprone-easily-swappable-parameters,
+    // readability-convert-member-functions-to-static)
+    {
+        // Sine Gordon 3D psuedo-breather solution
+
+        return 0.0;
```
mirenradia: Why is this 0?

Examples/KleinGordon/DerivedVariables.impl.hpp
```diff
@@ -0,0 +1,138 @@
...
+            amrex::Real y = problo[1] + (j + 0.5) * dx[1] - center[1];
+            amrex::Real z = problo[2] + (k + 0.5) * dx[2] - center[2];
+
+            arrs[box_no](i, j, k, dcomp) = model.calculate(x, y, z, time);
+            arrs[box_no](i, j, k, dcomp + 1) =
+                model.calculate_derivative(x, y, z, time);
```
mirenradia: Maybe it would be clearer if this function was called `calculate_time_derivative`?

Examples/KleinGordon/StateVariables.hpp
```diff
@@ -0,0 +1,31 @@
...
+    NUM_VARS,
+};
+
+namespace StateVariables
+{
+static const amrex::Vector<std::string> names{"phi", "Pi"};
+
+static const std::array<BCParity, NUM_VARS> parities = {BCParity::even,
+                                                        BCParity::even};
+// The parities aren't really used, I just had to fill something in
```
mirenradia: I think the wave solution is spherically symmetric so reflective BCs could be
used? I think the parities should be
```suggestion
static const std::array<BCParity, NUM_VARS> parities = {BCParity::even,
                                                        BCParity::odd};
```

Examples/KleinGordon/params_test.txt
```diff
@@ -0,0 +1,63 @@
...
+sigma = 0.3 #coefficient for KO dissipation
+
+tag_buffer_size=1
+
+#Klein Gordon Parameters
+# Choose from "Wave" or "SineGordon1D" or "SineGordon3D"
+# If "Wave" is specified then wave_vector must also be defined
+# If SineGordon is specified then alpha must also be defined
+# If SineGordon3D is specified then you also need an initial time
+# that can't be 0 because then the solution is uniformly 0
```
mirenradia: Might as well add the extra parameters needed in each case commented out.

Examples/KleinGordon/params_test.txt
```diff
@@ -0,0 +1,63 @@
...
+L_full              =  1     # geometric extent (same for all directions)
+N_full              =  64    # size of coarse grid
+
+
+#################################################
+# AMR parameters
+
+verbosity           =  1
+max_level           =  0     # maximum level number allowed
+regrid_interval     =  2 2 2 # how often to regrid
```
mirenradia: Is this needed if `max_level = 0`?

.pre-commit-config.yaml
mirenradia, line 1: To keep this PR atomic, can you move the changes here to a separate PR?

.gitlab-ci.yml
```diff
@@ -38,14 +46,29 @@ csd3-a100-test:
...
+      ${BINARY_DIR}/particle_compare.exe ${ERR_TOL_FLAGS}
+      ${OUTPUT_DIR}/plt00008/ ${COMPARE_DIR}/plt00008_BinaryBH_compare/ particles
+    - cd ${HOME}/${CI_PROJECT_DIR}/Examples/KleinGordon
+    - >  # Run the Klein Gordon example
+      flock ${HOME}/${QOS_INTR_LOCK_FILE}
+      srun ${SRUN_FLAGS} ${BINARY_DIR}/KleinGordon_main3d.gnu.DEBUG.MPI.CUDA.ex
+      ./params_test.txt amr.plot_file=${OUTPUT_DIR}/plt
+    - >  # Use fcompare to check for correctness
+      ${BINARY_DIR}/fcompare.gnu.ex ${ERR_TOL_FLAGS}
+      ${OUTPUT_DIR}/plt00008/ ${COMPARE_DIR}/plt00008_KleinGordon_compare
```
mirenradia: I think it's probably time to run the tests and the two regression tests
together in a single job (otherwise it has to wait in the queue 3 times) by moving some of these commands into a job submission
script. We can make the `sbatch` command blocking (until the job completes) with the `-W` flag.

mirenradia (changes requested):
OK, I have now gone through the files I didn't manage to review previously. Please see the individual comments.

Examples/KleinGordon/KleinGordonLevel.cpp
julianakwan: I just wanted any example of tagging really. I chose the fixed grid tagger as it is the most appropriate out of the other tagging criteria that have been ported so far.

I have added a comment to that effect: 
```
// This is an example of how to apply a tagging criterion to mark cells for                                            
// further refinement.                                                                                                 

// The FixedGridsTagger used here will always tag the inner                                                            
// L/4 cells of the simulation volume, regardless of the                                                               
// field values. You can also ask that the cells be tagged                                                             
// when the field value or its derivative reaches a certain value.  
```

Examples/KleinGordon/DerivedVariables.impl.hpp
julianakwan: I've made the suggested changes, but I would like to point out that the `Coordinates` class is actually less flexible than how I've written it now, because it assumes that `dx=dy=dz`

mirenradia: I think we make that assumption elsewhere in the code.

Examples/KleinGordon/params_test.txt
julianakwan: It has to be defined even if the run is coarse level only. Otherwise the run exits with:
```
ParmParse::sgetarr(): regrid_interval not found in database
```
I just set it to `regrid_interval = 2` for a single level.

Examples/KleinGordon/KleinGordonLevel.cpp
julianakwan: The lambda function doesn't work unfortunately:
```
KleinGordonLevel.cpp:122:13: error: return type 'SineGordon' must match previous return type 'Wave' when lambda expression has unspecified explicit return type
  122 |             return SineGordon{};
```
I think the only way to achieve something like this is the [factory method](https://www.geeksforgeeks.org/system-design/factory-method-pattern-c-design-patterns/). The reason I didn't implement it initially was because I think it reduces the readability of the code because of the added complexity.

Examples/KleinGordon/KleinGordonRHS.hpp
julianakwan: Since I still use the `Vars` structure in the RHS calculation and the `CellData` wrapper is still yet to be determined (e.g. I think it is quite likely that `CC4Vars` and `ConstCC4Vars` will be merged together as a single class somehow), I would like to leave it as it for now and replace it later when all changes have been incorporated into `develop`.

.pre-commit-config.yaml
julianakwan: This is now a new PR #142 

I have edited the Git log to remove mentions of the changes also.

.github/workflows/regression-tests-common.yml
julianakwan: I'm not quite sure what you had in mind... I let this be an environment variable because I thought we might change this to something like `$${ example_name }3d.$${ comp }.MPI.ex` one day. So I think it makes sense to leave it as a variable and then change it to incorporate the other variables when we've renamed the executable.

.gitlab-ci.yml
julianakwan: Thanks! I have something that works now but I think it could be tidier if the batch submission script accepted the name of the example as an input like this: 
```
sbatch -W --export=EXAMPLE_NAME=BinaryBH .gitlab/run-examples.sh
```
and then inside the script you would use that to run whichever example plus the tests with `fcompare` etc.
But this requires 2 job submissions that must be run serially and we may add more examples later. What do you think?

.github/workflows/regression-tests-common.yml
julianakwan: I changed it to `regression-tests-examples-template.yml` so as to not be confused with the composite action.

Examples/KleinGordon/params_test.txt
mirenradia: You're right. This is because of what we do in `AMReXParameters` and should be resolved when #1 is done. AMReX does this correctly for `regrid_int` (see [here](https://github.com/AMReX-Codes/amrex/blob/5d1b6c2bdec3498971fa4a00788e4bab3aeaaac2/Src/Amr/AMReX_Amr.cpp#L377-L413)).

Examples/KleinGordon/KleinGordonLevel.cpp
mirenradia: What about using `std::variant` and `std::visit` instead and doing something like the following:

```cpp
#include <variant>

...

    using ModelVariant = std::variant<Wave, SineGordon>;
    ModelVariant my_model_variant;
    if (model == "Wave")
    {
        my_model_variant = Wave{};
    }
    else
    {
        my_model_variant = SineGordon{};
    }

    std::visit(
        [&](auto &&my_model)
        {
            KleinGordonRHS rhs(simParams().sigma, dx, my_model);
            amrex::ParallelFor(
                a_soln,
                [=] AMREX_GPU_DEVICE(int box_no, int i, int j, int k) noexcept
                { rhs.compute(i, j, k, soln_arrs[box_no], rhs_arrs[box_no]); });
        },
        my_model_variant);
```

Examples/KleinGordon/KleinGordonRHS.hpp
mirenradia: Given there are only 2 variables, we can probably not bother with any "vars" structs at all and instead just load in the variables we need (and also only calculate the derivatives we need)?

.github/workflows/regression-tests-common.yml
mirenradia: I think you've now changed these workflow configurations significantly enough that my comment doesn't apply any more so I will resolve it.

julianakwan: I actually see what you mean now and we should change it use the matrix name when we get round to changing the  name of the executable one day. But yes, I don't know if there is anything I can do about it as it is.

Examples/KleinGordon/KleinGordonLevel.cpp
julianakwan: CUDA does not like your solution:
```
KleinGordonLevel.cpp(153): error: An extended __device__ lambda cannot be defined inside a generic lambda expression("operator()").
```
I had grown to like it myself and had even written a long comment explaining exactly what has happening 🥹

mirenradia: I guess it might be possible to get around this by changing the lambda argument to `std::visit` into a defined functor but perhaps now this is getting too complicated.

Perhaps a simpler solution is just to define a new template member function (which will need to be defined in the header) which is templated over the model and is something like:

```cpp
template <class model_t>
void eval_model_specific_rhs(amrex::MultiFab &a_soln, amrex::MultiFab &a_rhs)
{
    const auto dx         = Geom().CellSize(0);
    auto const &soln_arrs = a_soln.const_arrays();
    auto const &rhs_arrs  = a_rhs.arrays();

    model_t my_model;

    // I think we can use C++17 CTAD here so shouldn't need to explicitly
    // specify template arugments
    KleinGordonRHS klein_gordon_rhs(simParams().sigma, dx, my_model);

    amrex::ParallelFor(
        a_soln,
        [=] AMREX_GPU_DEVICE(int box_no, int i, int j, int k) noexcept
        {
            klein_gordon_rhs.compute(i, j, k, soln_arrs[box_no],
                                     rhs_arrs[box_no]);
        });
}
```
The first 3 lines here can be removed from `KleinGordonLevel::specificEvalRHS()` and the conditional block can be something like
```cpp
    pp.query("model", model);

    if (model == "Wave")
    {
        eval_model_specific_rhs<Wave>(a_soln, a_rhs);
    }
    else
    {
        eval_model_specific_rhs<SineGordon>(a_soln, a_rhs);
    }
```
NB. I haven't tested any of the code above so it could not work at all. Hopefully it's not too far off something that could work.

.github/workflows/regression-tests.yml
julianakwan: Sorry for being slow on this! I had actually forgotten why I had done it, but I remember now. It is because I didn't think the binary would compress very much for the extra time taken by the compression to be worth it. If you would prefer the default level of compression (6 by the way) I can change it. I think the default level assumes that it is a text file or something very compressible.

Examples/KleinGordon/KleinGordonLevel.cpp
julianakwan: I think writing the part inside the `std::visit` lambda as a functor would have worked but I ended up just writing another function inside `KleinGordonLevel` as you suggested.

Examples/KleinGordon/KleinGordonRHS.hpp
julianakwan: I have taken these out now and rewritten the derivatives although the code looks a bit messier now 😬

.gitlab-ci.yml
julianakwan: As discussed in our 1v1, this could be rewritten to use a single interactive session with`salloc` that runs a separate script for each example however, we would need to pass the Slurm job id for that session to the script. As this is a bit of a pain, we decided to shelve it for later development (when there are more examples).

Examples/KleinGordon/KleinGordonRHS.hpp
mirenradia: OK, it seems that there are some missing overloads in `FourthOrderDerivatives`. In particular, there is a `diff2` overload that calculates all second derivatives for a single variable with signature that Katy added in #127:
```cpp
    [[nodiscard]] AMREX_GPU_DEVICE AMREX_FORCE_INLINE Tensor<2, amrex::Real>
    diff2(int i, int j, int k, const amrex::Array4<amrex::Real const> &state,
          const int ivar) const
```
but not the equivalent for `diff1`, `add_dissipation` or `advection`.

I will add these overloads in a separate PR.

mirenradia: Actually I realise that @KAClough is already adding these overloads in #138 so I won't bother. Even though it is not a bit messier, I think I will just accept it for now and we can change it later.

Examples/KleinGordon/StateVariables.hpp
mirenradia: Actually, I think I'm wrong here. I got confused between spatial derivatives (which changes the parity) and time derivatives (which don't alter the parity). Could you change it back? Sorry about that :slightly_frowning_face:

.github/workflows/regression-tests.yml
mirenradia: I don't mind. Let's just leave it as it is.

Examples/KleinGordon/params_test.txt
```diff
@@ -0,0 +1,70 @@
...
+stop_time = 2
+
+#################################################
+# Geometry and Boundary Conditions parameters
+
+# Periodic directions - 0 = false, 1 = true
+isPeriodic          =  1   1   1
+center              =  0.5 0.5 0.5
+
+L_full              =  1     # geometric extent (same for all directions)
```
mirenradia: I ran this parameter file and thought the domain was rather small given the length scales of the solution. Maybe it's worth making it bigger? e.g. 
```suggestion
L_full              =  32.0     # geometric extent (same for all directions)
```

Examples/KleinGordon/DerivedVariables.hpp
mirenradia: OK

mirenradia:
Modulo the parity and `L_full` parameter comment, I'm happy to approve this PR. I think @KAClough wants to have a look before we merge.

KAClough:
Yes, I just had a look and it looks good! There are a few residual `data_t`s that I think could be got rid of? But we can also do this after. 

The main thing I wanted to request was a more explicit statement in the README about the fact that this doesn't have any GR. Maybe something like "Please note that this example does *not* have a dynamical metric, and simply assumes a Minkowski-flat background metric. It is intended to provide a simple example that implements a real scalar wave equation, showing how different initial data, diagnostics and parameters can be set."

julianakwan:
Hi @mirenradia and @KAClough - just letting you know that I have addressed all the comments now, including removing all the remaining instances of `data_t`, so this PR is ready for another review. Thanks!

mirenradia (approved):
LGTM

PR #123, opened 2025-06-23
CI: Add style check for all files

julianakwan:
This will add another GitHub workflow `style.yml` that checks for tabs and white spaces. 

`style.yml` calls two shell script in contained in `.github/workflows/style` - `check_trailing_whitespaces.sh` and `check_tabs.sh`. 

It will check any files ending in:
- *.F90, *.f90, *.f, *.F
- *.py
- *.md
- *.rst
- *.txt
- *.tex
- *.yml

I removed the checks for C++ related files because these should be covered under the existing linter. 

I purposely didn't fix the points that it picked up on so you can see [here](https://github.com/GRTLCollaboration/GRTeclyn/actions/runs/15831352090) how it outputs and the kind of things that it is picking up. (I will fix it later once if you choose to go with this option). I actually don't think the parameter files should be included, but I am okay if you choose to include these as well. 

NB: I have repurposed these from the AMReX repo, hence the branch name.

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -1,18 +1,18 @@
 name: Style
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   tabs:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Tabs
         run: .github/workflows/style/check_tabs.sh
 
   trailing_whitespaces:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Trailing Whitespaces
         run: .github/workflows/style/check_trailing_whitespaces.sh
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -1,0 +1,18 @@
+name: Style
+
+on: [push, pull_request]
+
+jobs:
+  tabs:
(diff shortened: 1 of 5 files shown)


mirenradia:
Thanks for this PR! It looks good to me.

> I actually don't think the parameter files should be included, but I am okay if you choose to include these as well.

Why not? I'm inclined to include them but if you think there's a good reason to exclude them, I don't feel too strongly.

julianakwan:
I left the text files to be checked as is and fixed the white spaces on the parameter files but I think our users may not like us very much 😄

.github/workflows/style.yml
```diff
@@ -0,0 +1,18 @@
+name: Style
+
+on: [push, pull_request]
```
mirenradia: It's probably sufficient just to run this workflow on `push`. I don't think it needs to be re-run when a PR is opened.
```suggestion
on: [push]
```

.github/workflows/style.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  tabs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
mirenradia:
```suggestion
      - uses: actions/checkout@v4
```

mirenradia (commented):
Just a couple of minor changes.

julianakwan:
Done! Thanks for the comments!

mirenradia (approved):
LGTM

PR #124, opened 2025-06-23
Pre-commit: Add hook for yaml file linting

julianakwan:
This PR adds [`yamllint`](https://yamllint.readthedocs.io/en/latest/) as a pre-commit hook. It offers a much more comprehensive check of `yaml` files than the style checks in `test/amrex-style-files`, but is also a bit annoying in that it will pick up on our excessively long lines and the weird indenting in `gpu-build.yml` (which I have already fixed). I changed the flag for long lines from an error to a warning, as some of those lines cannot be broken without affecting readability (as it will be a link for a download or some such).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,4 +14,4 @@ repos:
     rev: v1.29.0
     hooks:
       - id: yamllint
-
+        args: ['-d {rules: {line-length: {level: warning}}}']
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,4 +14,7 @@ repos:
     rev: v1.29.0
     hooks:
       - id: yamllint
-        args: ['-d {rules: {line-length: {level: warning}}}']
+        args: 
+          ['-d {rules: {line-length: {max: 101, allow-non-breakable-words: true,
+          allow-non-breakable-inline-mappings: true}}}']
+
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,17 @@
 repos:
--   repo: https://github.com/pre-commit/mirrors-clang-format
+  # Clang format
+  - repo: https://github.com/pre-commit/mirrors-clang-format
     rev: v19.1.7
     hooks:
-    -   id: clang-format
+      - id: clang-format
+  # Shell format
+  - repo: https://github.com/scop/pre-commit-shfmt
+    rev: v3.11.0-1
+    hooks:
+      - id: shfmt         # prebuilt upstream executable
+  # Yamllint    
+  - repo: https://github.com/adrienverge/yamllint.git
+    rev: v1.29.0
+    hooks:
+      - id: yamllint
 
-- repo: https://github.com/scop/pre-commit-shfmt
-  rev: v3.11.0-1
-  hooks:
-    - id: shfmt         # prebuilt upstream executable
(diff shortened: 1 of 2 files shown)


mirenradia:
Thanks for this PR! I'm happy to have this as well as #123. I don't think they're mutually exclusive.

For the line length, why don't we change it to max 80 chars and also set `allow-non-breakable-words` to allow for URLs? We could set it back to "error" in this case?

julianakwan:
> Thanks for this PR! I'm happy to have this as well as #123. I don't think they're mutually exclusive.
> 
> For the line length, why don't we change it to max 80 chars and also set `allow-non-breakable-words` to allow for URLs? We could set it back to "error" in this case?

I have changed the settings for `line-length` to allow for max 100 chars and set `allow-non-breakable-words` and `allow-non-breakable-inline-mappings` to `TRUE`. 

I think the readability of some lines have improved but also some lines are now _harder_ to read as well.

.pre-commit-config.yaml
```diff
@@ -1,10 +1,18 @@
...
+    rev: v3.11.0-1
+    hooks:
+      - id: shfmt         # prebuilt upstream executable
+  # Yamllint    
+  - repo: https://github.com/adrienverge/yamllint.git
+    rev: v1.29.0
+    hooks:
+      - id: yamllint
+        args: ['-d {rules: {line-length: {max: 101, allow-non-breakable-words: true,
+        allow-non-breakable-inline-mappings: true}}}']
```
mirenradia: Ironically, I think the YAML linter has added a line break here which I think breaks the intended configuration. Can you fix this?

julianakwan: I don't think it's broken as I wrote a short test case: 
```
this:
  is:
    a:
      test:
        - foo: just-a-really-really-long-line-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
`yamllint` will flag this if I set: 
```
	args: ['-d {rules: {line-length: {max: 101, allow-non-breakable-words: true,
	allow-non-breakable-inline-mappings: false}}}']
```
in `.pre-commit-config.yaml` so I think it is reading the second config option. 
[The one that really annoys me](https://github.com/GRTLCollaboration/GRTeclyn/blob/develop/.github/workflows/gpu-build.yml#L66) 
```
curl -O \
  https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/cuda-keyring_1.1-1_all.deb
```
cannot be broken sensibly unless the link is defined separately as a environment variable. There are a couple of other places that I can refactor in `gpu-build.yml` though - let me take another look

mirenradia: For legibility in the lines I've commented on, maybe it makes sense to use `>` to allow breaking the value of `args` over multiple lines but have the same indentation for those lines?

julianakwan: This is not allowed: 
```
        args: >
          ['-d {rules: {line-length: {max: 101, allow-non-breakable-words: true,
          allow-non-breakable-inline-mappings: true}}}']
```
But this is allowed:
```
        args:
          ['-d {rules: {line-length: {max: 101, allow-non-breakable-words: true,
          allow-non-breakable-inline-mappings: true}}}']
```
Would you prefer that? The first one gives a python error from the `pre-commit` package.

mirenradia: Yes, I think the second one is better than what is there currently.

julianakwan:
I think I have resolved all my issues and rebased onto `develop` again. Could you please take another look @mirenradia ? Thanks!

mirenradia (approved):
Looks good to me!

Issue #101, opened 2025-03-18
Reduce enforcement of algebraic constraints


mirenradia:
Currently `TraceARemoval` and ~`PositiveChiAndAlpha`~`PositiveChiAndLapse` (#99) get called a lot which is just because we were too lazy to remove them and see when it breaks.

We should figure out where they are actually needed and remove any superfluous calls.

Moved from #92.

KAClough:
Having tested this we in fact need to increase and not reduce the amount of enforcement of the constraints. We need to also impose that the determinant of the spatial metric is 1. The good news is that it seems to work as well using an additional damping term in the eoms (as in https://arxiv.org/pdf/1705.01071) as it does overwriting the values in the current TraceARemoval calls. So we can remove them and just have the RHS updates. I will close the previous PR on this as it is now out of date for this update. Thanks to @llibert94 for testing this.

PR #125, opened 2025-06-29
Remove algebraic constraint enforcement in specificEvalRHS

the-florist:
This commit addresses Issue #101  by removing the algebraic constraint enforcement functions i.e. TraceARemoval() and PositiveChiAndLapse() from the specificEvalRHS function in the BinaryBH example.

By systematically removing each call to these functions in the BinaryBH example, I determined that the call to TraceARemoval() and the calls to both TraceARemoval() and PositiveChiAndLapse() in _either_ specificEvalRHS or specificAdvance are required in order to prevent crash, up to 6 time steps on the coarsest level, using the model given by params.txt.

I chose to remove the TraceARemoval() and PositiveChiAndLapse() functions from specificEvalRHS because I believe this function is called on each sub-step, whereas it seems that specificAdvance may be called on each coarse time step only.

### DIFF
--- a/Examples/BinaryBH/BinaryBHLevel.cpp
+++ b/Examples/BinaryBH/BinaryBHLevel.cpp
@@ -143,8 +143,6 @@ void BinaryBHLevel::specificEvalRHS(amrex::MultiFab &a_soln,
                        {
                            amrex::CellData<amrex::Real> cell =
                                soln_arrs[box_no].cellData(i, j, k);
-                           TraceARemoval()(cell);
-                           PositiveChiAndLapse()(cell);
                        });
 
     // Calculate CCZ4 right hand side


KAClough:
Hey I just saw this, sorry I missed it completely! Great! Are you able to do a run to merger for both set ups, and compare the norm of the constraint violation? You can use my time on Cosma8 if you need it, or I can do the runs.

The other thing which I am not sure we discussed was whether to implement the trace removal dynamically, as Miguel does in MHDuet. It would be good to have a third run with this, for comparison, using the additional terms as in this paper: https://arxiv.org/pdf/1705.01071. This is likely to be less exact, but should mean that we do it all in the RHS, and could get rid of the TraceARemoval function completely, which would be nice.

julianakwan:
Perhaps you can also use the `fcompare` tool to check the difference between outputs with and without those functions. I also suggest running it forwards for a few more timesteps as well 😄

PR #127, opened 2025-07-12
Remove templating for explicit SIMD vectorization

### LINKED ISSUE #103: Remove templating over `data_t` and `simd` folder
Since the switch to AMReX, we no longer use this.

Before we say goodbye for good, we should check that auto-vectorization is good enough and performance (or at least the RHS computation) is comparable to GRChombo on CPUs.

KAClough:
Removed the `data_t` objects throughout the code as per #103. Mostly it is straightforward, and just involves removing or replacing with a `amrex::Real`, and making some header files into cpps for linking.

A few notes:

1. The `Vars` objects are templated over a data_t. This naively looks like it should be removed, but it is not only used for the grid double/vector templates but when we calculate the derivs we use it for the structures there - e.g. in second derivatives code we use `vars_t<data_t>` with data_t a `Tensor<2,double>` object. Ultimately we want to remove these Vars objects completely, but I propose to do this in a separate PR and leave them templated for now.

2. Mostly I fixed the linter errors but I added ignores for a few that I felt were unnecessary - we always call the radius squared `r2` and I don't want to change that even if the linter thinks it is too short. 

3. The tests use old versions of the code that rely on the simd code, so I moved the simd folder there from the Source files rather than deleting it for now. I did not change the tests because that would a bit defeat the point of them, but I would like to amend them in the future. Again maybe it makes sense to update them once all our refactoring is done, and we have a more stable version of the code.

4. In many places I made e.g. the x coord a `amex::Real` and left the y coord a `double`. I have edited the issue to say that we should unify this in another PR.

5. I wasn't sure what to do in `SetupFunctions.hpp` with outputting the simd length so I just commented it out :-) Suggestions welcome!

6. I only edited code that was used by the Tests or BBH example, so that it got compiled. Things like the Kerr initial data will need to be done once we add that example back.


Source/CCZ4/PositiveChiAndLapse.hpp
```diff
@@ -34,8 +33,8 @@ class PositiveChiAndLapse
         auto chi   = cell[c_chi];
         auto lapse = cell[c_lapse];
 
-        chi   = simd_max(chi, m_min_chi);
-        lapse = simd_max(lapse, m_min_lapse);
+        chi   = (chi < m_min_chi) ? m_min_chi : chi;
+        lapse = (lapse < m_min_lapse) ? m_min_lapse : chi;
```
KAClough: Oops. Will fix this in next update.

KAClough: Fixed

Source/BlackHoles/Make.package
```diff
@@ -1,8 +1,8 @@
 GRTECLYN_CEXE_headers += BHAMR.hpp \
-                         BinaryBHInitialData.hpp \
-                         BinaryBHInitialData.impl.hpp \
                          BoostedBHInitialData.hpp \
                          BoostedBHInitialData.impl.hpp \
+                         BinaryBHInitialData.hpp \
```
julianakwan: This is my fault 😭 Can you please reorder the files in alphabetical order again? Thanks!

KAClough: No worries! I will do it, enjoy your holiday!

mirenradia: When resolving this, please can you make sure the line indentation is all
spaces and no tabs (the line below currently has a mixture)?

Source/CCZ4/CCZ4Geometry.hpp
```diff
@@ -45,18 +43,16 @@ class CCZ4Geometry
...
   public:
-    template <class data_t, template <typename> class vars_t,
-              template <typename> class diff2_vars_t>
-    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE static ricci_t<data_t>
-    compute_ricci_Z(const vars_t<data_t> &vars,
-                    const vars_t<Tensor<1, data_t>> &d1,
-                    const diff2_vars_t<Tensor<2, data_t>> &d2,
-                    const Tensor<2, data_t> &h_UU, const chris_t<data_t> &chris,
-                    const Tensor<1, data_t> &Z_over_chi)
+    template <class vars_t, class d1_vars_t, class d2_vars_t>
```
mirenradia: Is it necessary to include the `d1_vars_t` template parameter? I think it can
inferred from `vars_t` since it's just `vars_t<Tensor<1, amrex::Real>>`?

Source/CCZ4/CCZ4Geometry.hpp
```diff
@@ -143,24 +139,22 @@ class CCZ4Geometry
...
     // to the Ricci scalar rather than the default of 2 in compute_ricci_Z
-    template <class data_t, template <typename> class vars_t,
-              template <typename> class diff2_vars_t>
-    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE static ricci_t<data_t>
-    compute_ricci_Z_general(const vars_t<data_t> &vars,
-                            const vars_t<Tensor<1, data_t>> &d1,
-                            const diff2_vars_t<Tensor<2, data_t>> &d2,
-                            const Tensor<2, data_t> &h_UU,
-                            const chris_t<data_t> &chris, const double dZ_coeff)
+    template <class vars_t, class d1_vars_t, class d2_vars_t>
```
mirenradia: As previously:
> Is it necessary to include the `d1_vars_t` template parameter? I think it can
inferred from `vars_t` since it's just `vars_t<Tensor<1, amrex::Real>>`?

Source/CCZ4/CCZ4Geometry.hpp
```diff
@@ -187,13 +182,10 @@ class CCZ4Geometry
...
     // This function returns the pure Ricci scalar with no contribution from the
     // Z vector - used e.g. in the constraint calculations.
-    template <class data_t, template <typename> class vars_t,
-              template <typename> class diff2_vars_t>
-    AMREX_GPU_DEVICE AMREX_FORCE_INLINE static ricci_t<data_t>
-    compute_ricci(const vars_t<data_t> &vars,
-                  const vars_t<Tensor<1, data_t>> &d1,
-                  const diff2_vars_t<Tensor<2, data_t>> &d2,
-                  const Tensor<2, data_t> &h_UU, const chris_t<data_t> &chris)
+    template <class vars_t, class d1_vars_t, class d2_vars_t>
```
mirenradia: As previously:
> Is it necessary to include the `d1_vars_t` template parameter? I think it can
inferred from `vars_t` since it's just `vars_t<Tensor<1, amrex::Real>>`?

Source/CCZ4/CCZ4RHS.hpp
```diff
@@ -104,18 +102,14 @@ class CCZ4RHS
      *at least the members: chi, h[i][j], Gamma[i], A[i][j], Theta, lapse and
      *shift[i].
      **/
-    template <class data_t, template <typename> class vars_t,
-              template <typename> class diff2_vars_t>
+    template <class vars_t, class d1_vars_t, class d2_vars_t>
```
mirenradia: As previously:
> Is it necessary to include the `d1_vars_t` template parameter? I think it can
inferred from `vars_t` since it's just `vars_t<Tensor<1, amrex::Real>>`?

Source/CCZ4/Constraints.hpp
```diff
@@ -96,17 +94,14 @@ class Constraints
     Interval m_c_Moms_abs_terms;
     double m_cosmological_constant;
 
-    template <class data_t, template <typename> class vars_t,
-              template <typename> class diff2_vars_t>
-    AMREX_GPU_DEVICE Vars<data_t> constraint_equations(
-        const vars_t<data_t> &vars, const vars_t<Tensor<1, data_t>> &d1,
-        const diff2_vars_t<Tensor<2, data_t>> &d2,
-        const Tensor<2, data_t> &h_UU, const chris_t<data_t> &chris) const;
+    template <class vars_t, class d1_vars_t, class d2_vars_t>
```
mirenradia: As previously:
> Is it necessary to include the `d1_vars_t` template parameter? I think it can
inferred from `vars_t` since it's just `vars_t<Tensor<1, amrex::Real>>`?

Source/CCZ4/GammaCalculator.hpp
```diff
@@ -38,12 +38,13 @@ class GammaCalculator
...
 
-    template <class data_t> void compute(Cell<data_t> current_cell) const
+    void compute(Cell current_cell) const
     {
         // copy data from chombo gridpoint into local variables, and calc 1st
         // derivs
-        const auto vars = current_cell.template load_vars<Vars>();
-        const auto d1   = m_deriv.template diff1<Vars>(current_cell);
+        Vars<amrex::Real> vars;
+        load_vars(current_cell, vars);
```
mirenradia, line 46: This code hasn't been ported yet so maybe we should add a comment that we've
just removed the templating
```suggestion
        // TODO: Port this code
        // This code hasn't been ported to GRTeclyn.
        // We've just removed templating over data_t.
        Vars<amrex::Real> vars = load_vars(current_cell);
```

Source/CCZ4/IntegratedMovingPunctureGauge.hpp
```diff
@@ -73,13 +73,10 @@ class IntegratedMovingPunctureGauge
         current_cell.store_vars(B, GRInterval<c_B1, c_B3>());
     }
 
-    template <class data_t, template <typename> class vars_t,
-              template <typename> class diff2_vars_t>
+    template <class vars_t, class d1_vars_t, class d2_vars_t>
```
mirenradia: As previously:
> Is it necessary to include the `d1_vars_t` template parameter? I think it can
inferred from `vars_t` since it's just `vars_t<Tensor<1, amrex::Real>>`?

Source/CCZ4/IntegratedMovingPunctureGauge.hpp
```diff
@@ -59,11 +59,11 @@ class IntegratedMovingPunctureGauge
     // fillAllGhosts();
     // BoxLoops::loop(IntegratedMovingPunctureGauge(m_p.ccz4_params),
     // m_state_new, m_state_new, EXCLUDE_GHOST_CELLS);
-    template <class data_t> void compute(Cell<data_t> current_cell) const
+    void compute(Cell<amrex::Real> current_cell) const
```
mirenradia, line 62:
```suggestion
    // TODO: Port this class
    // We've just removed templating over data_t
    void compute(Cell<amrex::Real> current_cell) const
```

Source/CCZ4/MovingPunctureGauge.hpp
```diff
@@ -43,14 +43,10 @@ class MovingPunctureGauge
 
   public:
     MovingPunctureGauge(const params_t &a_params) : m_params(a_params) {}
-
-    template <class data_t, template <typename> class vars_t,
-              template <typename> class diff2_vars_t>
+    template <class vars_t, class d1_vars_t, class d2_vars_t>
```
mirenradia: As previously:
> Is it necessary to include the `d1_vars_t` template parameter? I think it can
inferred from `vars_t` since it's just `vars_t<Tensor<1, amrex::Real>>`?

Source/CCZ4/PositiveChiAndLapse.hpp
```diff
@@ -27,15 +26,14 @@ class PositiveChiAndLapse
...
-    operator()(const amrex::CellData<data_t> &cell) const
+    operator()(const amrex::CellData<amrex::Real> &cell) const
     {
         auto chi   = cell[c_chi];
         auto lapse = cell[c_lapse];
 
-        chi   = simd_max(chi, m_min_chi);
-        lapse = simd_max(lapse, m_min_lapse);
+        chi   = (chi < m_min_chi) ? m_min_chi : chi;
+        lapse = (lapse < m_min_lapse) ? m_min_lapse : lapse;
```
mirenradia:
```suggestion
        chi   = std::max(chi, m_min_chi);
        lapse = std::max(lapse, m_min_lapse);
```

Source/CCZ4/TraceARemoval.hpp
```diff
@@ -17,21 +17,22 @@
...
 
-    template <class data_t>
     AMREX_GPU_HOST_DEVICE void
-    operator()(const amrex::CellData<data_t> &cell) const
+    operator()(const amrex::CellData<amrex::Real> &cell) const
     {
-        auto vars = load_vars<Vars>(cell);
+        // NOLINTNEXTLINE [cppcoreguidelines-pro-type-member-init]
+        Vars vars;
+        load_vars(cell, vars);
```
mirenradia: Does the following work?
```suggestion
        Vars vars = load_vars<Vars>(cell);
```

Source/CCZ4/Weyl4.impl.hpp
```diff
@@ -259,7 +254,7 @@ Weyl4::compute_null_tetrad(const Vars<data_t> &vars,
     out.w[2] = 0.0;
 
     // floor on chi
-    const data_t chi = simd_max(vars.chi, 1e-4);
+    const amrex::Real chi = (vars.chi < 1e-4) ? 1e-4 : vars.chi;
```
mirenradia: Maybe this should use the same same `min_chi` as used in `PositiveChiAndLapse`
but that's for another time. In any case:
```suggestion
    const amrex::Real chi = std::max(vars.chi, 1.0e-4_rt);
```

Source/GRTeclynCore/SetupFunctions.hpp
```diff
@@ -36,7 +34,8 @@ void mainSetup(int argc, char *argv[]);
 void mainFinalize();
 
 #if !defined(AMREX_USE_GPU)
-const int simd_traits<double>::simd_len; // Still needs to be defined
+// TODO: FIXME
+// const int simd_traits<double>::simd_len; // Still needs to be defined
 #endif
```
mirenradia: I think we can just get rid of this:
```suggestion
```

Source/GRTeclynCore/SetupFunctions.hpp
```diff
@@ -53,8 +52,9 @@ void mainSetup(int argc, char *argv[])
...
 
 #if !defined(AMREX_USE_GPU)
-    amrex::Print() << " simd width (doubles) = "
-                   << simd_traits<double>::simd_len << '\n';
+// TODO: FIXME
+//    amrex::Print() << " simd width (doubles) = "
+//                   << simd_traits<double>::simd_len << '\n';
 #endif
```
mirenradia: And this!
```suggestion
```

Source/Grids/Cell.hpp
```diff
@@ -10,18 +10,27 @@
...
         });
 }
 
+template <class vars_t>
+AMREX_GPU_HOST_DEVICE void load_vars(const amrex::CellData<amrex::Real> &cell,
+                                     vars_t &vars)
+{
+    vars.enum_mapping([&](const int &ivar, amrex::Real &var)
+                      { var = cell[ivar]; });
+}
```
mirenradia: Why did you create this new template function rather than just modifying the
template functions below accordingly?

Source/Grids/Coordinates.hpp
```diff
@@ -45,65 +44,48 @@ template <class data_t> class Coordinates
...
     get_radius() const
     {
         // Note that this is not currently dimension independent
-        data_t r = sqrt(x * x + y * y + z * z);
+        amrex::Real r = sqrt(x * x + y * y + z * z);
 
-        const double minimum_r = 1e-6;
-        return simd_max(r, minimum_r);
+        const amrex::Real minimum_r = 1e-6;
+        return (r < minimum_r) ? minimum_r : r;
```
mirenradia:
```suggestion
        return std::max(r, minimum_r);
```

Source/Grids/Coordinates.hpp
```diff
@@ -45,65 +44,48 @@ template <class data_t> class Coordinates
...
         compute_coord(y, integer_coords[1], dx, center[1]);
         compute_coord(z, integer_coords[2], dx, center[2]);
 
-        data_t r = std::sqrt(x * x + y * y + z * z);
+        amrex::Real r = std::sqrt(x * x + y * y + z * z);
 
-        const double minimum_r = 1e-6;
-        return simd_max(r, minimum_r);
+        const amrex::Real minimum_r = 1e-6;
+        return (r < minimum_r) ? minimum_r : r;
```
mirenradia:
```suggestion
        return std::max(r, minimum_r);
```

Source/CCZ4/Weyl4.impl.hpp
```diff
@@ -229,18 +226,16 @@ AMREX_GPU_DEVICE AMREX_FORCE_INLINE NPScalar_t<data_t> Weyl4::compute_Weyl4(
...
-    Tetrad_t<data_t> out;
+    Tetrad_t out;
 
     // compute coords
-    const data_t x = coords.x;
-    const double y = coords.y;
-    const double z = coords.z;
+    const amrex::Real x = coords.x;
+    const double y      = coords.y;
+    const double z      = coords.z;
```
mirenradia: Given you changed it in `Coordinates.hpp`:
```suggestion
    const amrex::Real x = coords.x;
    const amrex::Real y = coords.y;
    const amrex::Real z = coords.z;
```

Source/Grids/FourthOrderDerivatives.hpp
```diff
@@ -9,142 +9,149 @@
...
-    template <class data_t>
-    AMREX_GPU_DEVICE AMREX_FORCE_INLINE data_t diff1(const double *in_ptr,
-                                                     const int idx,
-                                                     const int stride) const
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE amrex::Real
+    diff1(const amrex::Real *in_ptr, const int idx, const int stride) const
     {
-        const auto *in = SIMDIFY<data_t>(in_ptr);
+        amrex::Real weight_far  = 8.33333333333333333333e-2;
+        amrex::Real weight_near = 6.66666666666666666667e-1;
```
mirenradia:
```suggestion
        amrex::Real weight_far  = 8.33333333333333333333e-2_rt;
        amrex::Real weight_near = 6.66666666666666666667e-1_rt;
```

Source/Grids/FourthOrderDerivatives.hpp
```diff
@@ -9,142 +9,149 @@
...
+    diff2(const amrex::Real *in_ptr, const int idx, const int stride) const
     {
-        const auto *in = SIMDIFY<data_t>(in_ptr);
-
-        data_t weight_far   = 8.33333333333333333333e-2;
-        data_t weight_near  = 1.33333333333333333333e+0;
-        data_t weight_local = 2.50000000000000000000e+0;
+        amrex::Real weight_far   = 8.33333333333333333333e-2;
+        amrex::Real weight_near  = 1.33333333333333333333e+0;
+        amrex::Real weight_local = 2.50000000000000000000e+0;
```
mirenradia:
```suggestion
        amrex::Real weight_far   = 8.33333333333333333333e-2_rt;
        amrex::Real weight_near  = 1.33333333333333333333e+0_rt;
        amrex::Real weight_local = 2.50000000000000000000e+0_rt;
```

Source/Grids/FourthOrderDerivatives.hpp
```diff
@@ -9,142 +9,149 @@
...
-    AMREX_GPU_DEVICE AMREX_FORCE_INLINE data_t
-    mixed_diff2(const double *in_ptr, const int idx, const int stride1,
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE amrex::Real
+    mixed_diff2(const amrex::Real *in_ptr, const int idx, const int stride1,
                 const int stride2) const
     {
-        const auto *in = SIMDIFY<data_t>(in_ptr);
+        amrex::Real weight_far_far   = 6.94444444444444444444e-3;
+        amrex::Real weight_near_far  = 5.55555555555555555556e-2;
+        amrex::Real weight_near_near = 4.44444444444444444444e-1;
```
mirenradia:
```suggestion
        amrex::Real weight_far_far   = 6.94444444444444444444e-3_rt;
        amrex::Real weight_near_far  = 5.55555555555555555556e-2_rt;
        amrex::Real weight_near_near = 4.44444444444444444444e-1_rt;
```

Source/Grids/FourthOrderDerivatives.hpp
```diff
@@ -155,107 +162,106 @@ class FourthOrderDerivatives
...
-        data_t weight_0 = -2.50000000000000000000e-1;
-        data_t weight_1 = -8.33333333333333333333e-1;
-        data_t weight_2 = +1.50000000000000000000e+0;
-        data_t weight_3 = -5.00000000000000000000e-1;
-        data_t weight_4 = +8.33333333333333333333e-2;
+        amrex::Real weight_0 = -2.50000000000000000000e-1;
+        amrex::Real weight_1 = -8.33333333333333333333e-1;
+        amrex::Real weight_2 = +1.50000000000000000000e+0;
+        amrex::Real weight_3 = -5.00000000000000000000e-1;
+        amrex::Real weight_4 = +8.33333333333333333333e-2;
```
mirenradia:
```suggestion
        amrex::Real weight_0 = -2.50000000000000000000e-1_rt;
        amrex::Real weight_1 = -8.33333333333333333333e-1_rt;
        amrex::Real weight_2 = +1.50000000000000000000e+0_rt;
        amrex::Real weight_3 = -5.00000000000000000000e-1_rt;
        amrex::Real weight_4 = +8.33333333333333333333e-2_rt;
```

Source/Grids/FourthOrderDerivatives.hpp
```diff
@@ -155,107 +162,106 @@ class FourthOrderDerivatives
...
     {
-        const auto *const in = SIMDIFY<data_t>(in_ptr);
-        data_t weight_vfar   = 1.56250e-2;
-        data_t weight_far    = 9.37500e-2;
-        data_t weight_near   = 2.34375e-1;
-        data_t weight_local  = 3.12500e-1;
+        amrex::Real weight_vfar  = 1.56250e-2;
+        amrex::Real weight_far   = 9.37500e-2;
+        amrex::Real weight_near  = 2.34375e-1;
+        amrex::Real weight_local = 3.12500e-1;
```
mirenradia:
```suggestion
        amrex::Real weight_vfar  = 1.56250e-2_rt;
        amrex::Real weight_far   = 9.37500e-2_rt;
        amrex::Real weight_near  = 2.34375e-1_rt;
        amrex::Real weight_local = 3.12500e-1_rt;
```

Source/Maths/CoordinateTransformations.hpp
```diff
@@ -9,29 +9,33 @@
...
     // calculate useful position quantities
-    data_t rho2 = simd_max(x * x + y * y, 1e-12);
-    data_t rho  = sqrt(rho2);
-    // NOLINTNEXTLINE(readability-identifier-length)
-    data_t r2 = simd_max(x * x + y * y + z * z, 1e-12);
-    data_t r  = sqrt(r2);
+    amrex::Real rho2    = x * x + y * y;
+    amrex::Real rho     = std::sqrt(rho2);
+    amrex::Real min_rho = 1e-6;
+    rho                 = (rho < min_rho) ? min_rho : rho;
```
mirenradia:
```suggestion
    rho                 = std::max(rho, min_rho);
```

Source/Maths/CoordinateTransformations.hpp
```diff
@@ -9,29 +9,33 @@
...
+    // NOLINTBEGIN [readability-identifier-length]
+    amrex::Real r2 = x * x + y * y + z * z;
+    amrex::Real r  = std::sqrt(r2);
+    // NOLINTEND [readability-identifier-length]
+    amrex::Real min_r = 1e-6;
+    r                 = (r < min_r) ? min_r : r;
```
mirenradia:
```suggestion
    r                 = std::max(r, min_r);
```

Source/Maths/CoordinateTransformations.hpp
```diff
@@ -44,25 +48,29 @@ static Tensor<2, data_t> spherical_jacobian(const data_t x, const double y,
...
-    data_t rho2 = simd_max(x * x + y * y, 1e-12);
-    data_t rho  = sqrt(rho2);
-    // NOLINTNEXTLINE(readability-identifier-length)
-    data_t r2 = simd_max(x * x + y * y + z * z, 1e-12);
-    data_t r  = sqrt(r2);
+    // NOLINTBEGIN [readability-identifier-length]
+    amrex::Real rho2     = x * x + y * y;
+    amrex::Real min_rho2 = 1e-12;
+    rho2                 = (rho2 < min_rho2) ? min_rho2 : rho2;
+    amrex::Real rho      = std::sqrt(rho2);
```
mirenradia: For consistency with the previous function:
```suggestion
    amrex::Real rho     = std::sqrt(rho2)
    amrex::Real min_rho = 1e-6;
    rho                 = std::max(1e-6, rho);
```

Source/Maths/CoordinateTransformations.hpp
```diff
@@ -44,25 +48,29 @@ static Tensor<2, data_t> spherical_jacobian(const data_t x, const double y,
...
+    amrex::Real min_r2   = 1e-12;
+    amrex::Real r2       = x * x + y * y + z * z;
+    r2                   = (r2 < min_r2) ? min_r2 : r2;
+    amrex::Real r        = std::sqrt(r2);
```
mirenradia: For consistency with the previous function:
```suggestion
    amrex::Real r2      = x * x + y * y + z * z;
    amrex::Real r       = std::sqrt(r2);
    amrex::Real min_r   = 1e-6;
    r                   = std::max(r, min_r);
```

Source/Maths/SphericalHarmonics.hpp
```diff
@@ -35,7 +35,9 @@ spin_Y_lm(const data_t x, const double y, const double z,
     Y_lm_t<data_t> Y_lm{};
 
     // calculate useful position quantities
-    data_t r     = simd_max(sqrt(x * x + y * y + z * z), 1e-6);
+    data_t min_r = 1e-6;
+    data_t r     = sqrt(x * x + y * y + z * z);
+    r            = (r < min_r) ? min_r : r;
```
mirenradia:
```suggestion
    r            = std::max(r, min_r);
```

Source/Matter/CCZ4RHSWithMatter.hpp
mirenradia, line 1: You've missed most of the templating in this class! This code is tested in the BSSNMatterTest.

Source/Matter/ScalarField.hpp
```diff
@@ -73,26 +73,24 @@ template <class potential_t = DefaultPotential> class ScalarField
 
     //! The function which calculates the EM Tensor, given the vars and
     //! derivatives, including the potential
-    template <class data_t, template <typename> class vars_t>
-    AMREX_GPU_DEVICE emtensor_t<data_t> compute_emtensor(
-        const vars_t<data_t> &vars,          //!< the value of the variables
-        const vars_t<Tensor<1, data_t>> &d1, //!< the value of the 1st derivs
-        const Tensor<2, data_t> &h_UU, //!< the inverse metric (raised indices)
-        const Tensor<3, data_t> &chris_ULL)
+    template <class vars_t, class d1_vars_t>
```
mirenradia: As previously:
> Is it necessary to include the `d1_vars_t` template parameter? I think it can
inferred from `vars_t` since it's just `vars_t<Tensor<1, amrex::Real>>`?

Source/Matter/ScalarField.hpp
```diff
@@ -73,26 +73,24 @@ template <class potential_t = DefaultPotential> class ScalarField
...
+                         &h_UU, //!< the inverse metric (raised indices)
+                     const Tensor<3, amrex::Real> &chris_ULL)
         const; //!< the conformal christoffel symbol
 
     //! The function which adds in the RHS for the matter field vars,
     //! including the potential
-    template <class data_t, template <typename> class vars_t,
-              template <typename> class diff2_vars_t,
-              template <typename> class rhs_vars_t>
+    template <class rhs_vars_t, class vars_t, class d1_vars_t, class d2_vars_t>
```
mirenradia: As previously:
> Is it necessary to include the `d1_vars_t` template parameter? I think it can
inferred from `vars_t` since it's just `vars_t<Tensor<1, amrex::Real>>`?

Tests/CoordinateTransformationsTest/CoordinateTransformationsTest.cpp
mirenradia, line 1: Can you change the `double`s in this to `amrex::Real`s according to the changes
in `CoordinateTransformations.hpp`?

mirenradia (changes requested):
Thanks for the PR Katy. 

Some notes on my specific changes:
1. Many of my comments are on the introduction of the additional `d1_vars_t` template parameter which I don't believe to be necessary. I presume you did this to improve readability but this in fact further complicates the templating of these functions (thus slowing down compilation). I would be happy to instead use a type alias template e.g.
   ```cpp
   template <class vars_t>
   using d1_vars_t = vars_t<Tensor<1, amrex::Real>>;
   ```
1. I personally find the condition operator less readable than using `std::max`. Not sure why you seem to prefer the former.
1. When declaring a literal `amrex::Real`, we should use the `_rt` literal suffix (see [this part of the AMReX docs](https://amrex-codes.github.io/amrex/docs_html/Basics.html#real)) which is like the `f` suffix for declaring literal `float`s to avoid unnecessary type conversions (if we ever decide to compile with `amrex::Real = float`).


In addition to specific changes, I have a number of other suggestions:
1. Can you make `amrex::Real` the default value of the `data_t` template argument for the `Tensor` class template i.e.
   ```cpp
   template <int rank, class data_t = amrex::Real, int size = DEFAULT_TENSOR_DIM> class Tensor
   ```
   We can then remove `amrex::Real` when declaring a `Tensor` object e.g. `Tensor<2>` instead of `Tensor<2, amrex::Real>`.
1. I think the only test which _needs_ the SIMD code is the `CCZ4RHSTest` since it uses an old version of the `CCZ4RHS` class. Can you remove all references to it in other tests and then move this `simd` directory under the `CCZ4RHSTest` directory? I don't think modifying these tests in this simple way defeats the point of them.

Specifically for your point 2., rather than adding individual linter ignore lines for `r2`, can you instead add it the ` readability-identifier-length.IgnoredVariableNames` key in `.clang-tidy`?

Source/CCZ4/Weyl4.impl.hpp
KAClough: Actually I will just remove this. It should be unnecessary as it is enforced in PositiveChiAndLapse as you say, and then there won't be any inconsistency.

Source/Maths/CoordinateTransformations.hpp
KAClough: Actually this was done for a reason, because in one function rho2 and r2 are used as divisors, whereas in the other they are not. However, I got them the wrong way around! So they are not consistent, but I think they are now right!

Source/Maths/CoordinateTransformations.hpp
KAClough: As noted above.

Source/Matter/CCZ4RHSWithMatter.hpp
KAClough: :cry:

KAClough:
> Thanks for the PR Katy.

You are welcome!

> Some notes on my specific changes:
> 
> 1. Many of my comments are on the introduction of the additional `d1_vars_t` template parameter which I don't believe to be necessary. I presume you did this to improve readability but this in fact further complicates the templating of these functions (thus slowing down compilation). I would be happy to instead use a type alias template e.g.
>    ```c++
>    template <class vars_t>
>    using d1_vars_t = vars_t<Tensor<1, amrex::Real>>;
>    ```

Ha! You overestimate me. I did this because I couldn't figure out the syntax for vars_t itself being a template. It's actually a bit of a palaver, because you have to do `template <template <class> class vars_t>` and then make it `vars_t<amrex::Real>` in other places. But I did it, just to make you happy. I am going to delete it all in the next pull request anyway 🤣 

> 2. I personally find the condition operator less readable than using `std::max`. Not sure why you seem to prefer the former.

I thought it was what real coders did 😢 

> 3. When declaring a literal `amrex::Real`, we should use the `_rt` literal suffix (see [this part of the AMReX docs](https://amrex-codes.github.io/amrex/docs_html/Basics.html#real)) which is like the `f` suffix for declaring literal `float`s to avoid unnecessary type conversions (if we ever decide to compile with `amrex::Real = float`).

Ok 🫡 

> In addition to specific changes, I have a number of other suggestions:
> 
> 1. Can you make `amrex::Real` the default value of the `data_t` template argument for the `Tensor` class template i.e.
>    ```c++
>    template <int rank, class data_t = amrex::Real, int size = DEFAULT_TENSOR_DIM> class Tensor
>    ``` 
>      
>    We can then remove `amrex::Real` when declaring a `Tensor` object e.g. `Tensor<2>` instead of `Tensor<2, amrex::Real>`.

This is a good idea and I promise to do it in the next pull request for the vars. But I am already going to have to fix a lot of conflicts when I rebase to this and this will make it impossible, so I would prefer not to now.

> 2. I think the only test which _needs_ the SIMD code is the `CCZ4RHSTest` since it uses an old version of the `CCZ4RHS` class. Can you remove all references to it in other tests and then move this `simd` directory under the `CCZ4RHSTest` directory? I don't think modifying these tests in this simple way defeats the point of them.

Yes, I just meant we don't want to change the old files from the specific commit because the point is to compare to them. I moved it and I don't think any of the other files actually needed to change.

> Specifically for your point 2., rather than adding individual linter ignore lines for `r2`, can you instead add it the ` readability-identifier-length.IgnoredVariableNames` key in `.clang-tidy`?

Yes. I didn't know such a power existed... with great power comes great responsibility...

Source/CCZ4/TraceARemoval.hpp
KAClough: No it doesn't. I will get rid of it in the next update anyway...

Source/Grids/Cell.hpp
KAClough: Probably just due to desperation. I think these were still needed in the old form somewhere because of the vars templating still being a thing... I'll be able to get rid of all these horrible load and store functions soon.

Source/Matter/CCZ4RHSWithMatter.hpp
KAClough: Done, note that all the vars stuff has to keep its `data_t` s as noted above.

mirenradia:
> I thought it was what real coders did 😢

I would argue it's less C++-y!

> This is a good idea and I promise to do it in the next pull request for the vars. But I am already going to have to fix a lot of conflicts when I rebase to this and this will make it impossible, so I would prefer not to now.

I think it would be best to do it either here or in a separate PR outside the refactor vars PR. If you wish to defer to a separate PR, can you create an issue for it?

> Yes, I just meant we don't want to change the old files from the specific commit because the point is to compare to them. I moved it and I don't think any of the other files actually needed to change.

I think it's just the CCZ4RHSTest which references a specific commit. Can you remove the templating in the `HarmonicTest` class?

Source/Grids/Cell.hpp
mirenradia: I think it's just because you needed to create one also with `const amrex::CellData<const amrex::Real>&` for the type of `cell` (note the extra `const`).

If we were going to keep the vars, I would suggest either using [SFINAE](https://www.geeksforgeeks.org/cpp/substitution-failure-is-not-an-error-sfinae-in-cpp/) (my preference) or explicitly defining the two overloads (i.e. for `amrex::Real` and `const amrex::Real`

For now, for the sake of simplicity, maybe we can just remove the extra overload you have defined and rely on the ones templated over `data_t`?

Source/CCZ4/TraceARemoval.hpp
KAClough: Ok I fixed it.

Source/Grids/Cell.hpp
KAClough: Ok

Tests/CoordinateTransformationsTest/CoordinateTransformationsTest.cpp
```diff
@@ -29,8 +29,8 @@ namespace
 constexpr int ulp               = 15; // units in the last place
 constexpr double double_epsilon = std::numeric_limits<double>::epsilon();
```
mirenradia: Can you replace the `double` here with `amrex::Real` and also rename `double_epsilon` -> `real_epsilon`?

KAClough: Done

mirenradia (changes requested):
I think this PR is almost ready.

Before I approve, please could you interactively rebase this branch to clean the history. In particular, please can you

* Squash any "fix linter" or "wrong clang-format version" commits into earlier commits.
* Squash the final two commits.

Source/BlackHoles/BoostedBHInitialData.impl.hpp
```diff
@@ -59,46 +57,44 @@ BoostedBHInitialData::Aij(Coordinates<data_t> a_coords) const
...
-    data_t r = std::sqrt(std::pow(a_coords.x - m_params.center[0], 2) +
-                         std::pow(a_coords.y - m_params.center[1], 2) +
-                         std::pow(a_coords.z - m_params.center[2], 2));
+    amrex::Real r = std::sqrt(std::pow(a_coords.x - m_params.center[0], 2) +
+                              std::pow(a_coords.y - m_params.center[1], 2) +
+                              std::pow(a_coords.z - m_params.center[2], 2));
 
     double minimum_r = 1e-6;
-    return simd_max(r, minimum_r);
+    return (r < minimum_r) ? minimum_r : r;
```
mirenradia: I missed this one. Sorry 🙁 
```suggestion
    std::max(r, minimum_r);
```

Source/BlackHoles/Make.package
```diff
@@ -1,6 +1,6 @@
 GRTECLYN_CEXE_headers += BHAMR.hpp \
                          BinaryBHInitialData.hpp \
-                         BinaryBHInitialData.impl.hpp \
+         	               BinaryBHInitialData.impl.hpp \
```
mirenradia: There is a rogue tab on this line and I don't think this file needs to change.

Tests/CCZ4RHSTest/Make.package
mirenradia, line 1: Annoyingly, this file uses tabs rather than spaces. Could you switch it all to spaces?

.lint-ignore
```diff
@@ -14,10 +14,12 @@ Source/Grids/SixthOrderDerivatives.hpp
 Source/CCZ4/GammaCalculator.hpp
 Source/CCZ4/IntegratedMovingPunctureGauge.hpp
 Source/BlackHoles/KerrBH*
-Source/simd/*
 Source/IO/MultiLevelTask.hpp
 Examples/ScalarField/*
 Tests/.*
+Tests/CCZ4RHSTest/x64/*
```
mirenradia: Can this line be moved to the end of the file?

mirenradia (approved):
Looks good to me!

Issue #111, opened 2025-04-07
Replace loading and storing of variables in CCZ4RHS


julianakwan:
@WeiqunZhang pointed out that we were excessively loading and replacing the variables that we track on the grid in `CCZ4RHS::compute`. Neither [`load_vars`](https://github.com/GRTLCollaboration/GRTeclyn/blob/develop/Source/CCZ4/CCZ4RHS.impl.hpp#L49) or [`store_vars`](https://github.com/GRTLCollaboration/GRTeclyn/blob/develop/Source/CCZ4/CCZ4RHS.impl.hpp#L60) are strictly necessary, as `amrex::CellData` contains all of the relevant information packed in a GPU memory friendly way as a [1D array](https://github.com/AMReX-Codes/amrex/blob/development/Src/Base/AMReX_Array4.H#L55) with a stride. 

@KAClough, @the-florist and I started working on this during the AMD Hackathon (April 2 - April 4, 2025). The branch is called [`enhancement/vars_removal`](https://github.com/GRTLCollaboration/GRTeclyn/tree/enhancement/vars_removal)

We're currently seeing speedups of up to 40% on the MI300A with no reference to `load_vars` or `store_vars` but some `Tensor` objects are still being created (e.g [here](https://github.com/GRTLCollaboration/GRTeclyn/blob/enhancement/vars_removal/Source/CCZ4/CCZ4RHS.impl.hpp#L58) and [here](https://github.com/GRTLCollaboration/GRTeclyn/blob/enhancement/vars_removal/Source/CCZ4/CCZ4RHS.impl.hpp#L64) and I think they could also be replaced with 1D arrays. 

Still to do: 
- [ ] Rewrite tensor algebra functions to accept an unravelled 1D array of values 
- [ ] Symmetric indices are currently implemented in a hacky way [here](https://github.com/GRTLCollaboration/GRTeclyn/blob/enhancement/vars_removal/Source/utils/Macros.hpp#L43)
- [x] Fix the SYCL backend. Because of above hacky definition, the SYCL build is broken
- [ ] Give [`Macros.hpp`](https://github.com/GRTLCollaboration/GRTeclyn/blob/enhancement/vars_removal/Source/utils/Macros.hpp) a sensible name. 
- [ ] Fix all the broken regression and unit tests. 
[...]

PR #137, opened 2025-08-18
Removing the Vars objects that are created so we directly access grid data

KAClough:
This is in progress, but comments welcome. It addresses issue #111. 

The main goal is to remove all of the vars objects and replace them with a wrapper that directly accesses the right grid data. This involves removing all the enum mapping things, and also rethinking how the derivatives work. 

I am also trying to unify the way in which we use the Parallel4 loop, so that for users the examples are more easily copyable. For this reason I try to have it always call an operator() function of the class and to always pass in the full state and derive the cell data within the class (rather than passing in the cell data directly, which would then mean that you can't calculate derivatives for example). So it also addresses #8.


mirenradia:
Can you rebase this onto the current `develop` branch? There are commits from earlier iterations of #127 which are leading to the merge conflicts and making the diff bigger (it's already very big).

KAClough:
> Can you rebase this onto the current `develop` branch? There are commits from earlier iterations of #127 which are leading to the merge conflicts and making the diff bigger (it's already very big).

Give me a chance! You only just merged that!

Issue #128, opened 2025-07-21
Change `State_Type` to something more meaningful


mirenradia:
> Maybe we can also change the name `State_Type` to something less cryptic. Like `Temporary_State`? Or `Single_State`? Or just write above `base_index = 0` and then pass `base_index`? It's a very confusing way to provide the number 0. 

 _Originally posted by @KAClough in [#98](https://github.com/GRTLCollaboration/GRTeclyn/issues/98#issuecomment-3065906657)_

mirenradia:
I agree that it's not the clearest variable name. I think it's a name Weiqun chose. In theory, AMReX supports multiple states with different grids and/or index type (e.g. node/cell centered). We only use one type hence the need to pass `0` as the state index to many AMReX functions.

What about calling it something like `state_index_zero`?

KAClough:
Ok, maybe `zero_stste_index`?

mirenradia:
> Ok, maybe `zero_stste_index`?

Sounds good (assuming it's a typo and you meant `zero_state_index`).

KAClough:
This is done in PR #127, changing it to the more general `state_index` as suggested by @julianakwan

mirenradia:
> This is done in PR [#127](https://github.com/GRTLCollaboration/GRTeclyn/pull/127), changing it to the more general `state_index` as suggested by [@julianakwan](https://github.com/julianakwan)

I don't think it is (at least I can't see this change). I would also prefer that this is not done in #127 since it's an unrelated change and we should try and keep our PRs atomic (e.g. see [this article](https://medium.com/@chikim79/the-art-of-creating-atomic-prs-16d13b3828e3)). Lastly, I don't think we should use `state_index = 0` since it's possible in future we might introduce additional states. I would prefer to include `zero` in the name.

KAClough:
Ah yes sorry, it is not in #127 it is in what I am doing now. It does kind of make sense there as I am tidying up the whole interface of the ParallelFor thing... I am not sure that a link to a Medium article is sufficiently persuasive to make me change it back.

Let's discuss about the zero thing in the meeting tomorrow. If there are going to be other states, what might they be?  This might help us pick a less random name. Should it be `evolution_state` like we had in GRChombo, with the possible addition being `diagnostic_states`?

mirenradia:
> I am not sure that a link to a Medium article is sufficiently persuasive to make me change it back.

This was more to provide an explanation of the value of atomic PRs.

[...]

PR #138, opened 2025-08-18
Refactor vars, make compute classes consistent and fix some variable names

### LINKED ISSUE #111: Replace loading and storing of variables in CCZ4RHS

KAClough:
This addresses issue #111.

The main goal is to remove all of the vars objects and replace them with a wrapper that directly accesses the right grid data. This involves removing all the enum mapping things, and also rethinking how the derivatives work.

I am also trying to unify the way in which we use the Parallel4 loop, so that for users the examples are more easily copyable. For this reason I try to have it always call an operator() function of the class and to always pass in the full state and derive the cell data within the class (rather than passing in the cell data directly, which would then mean that you can't calculate derivatives for example). So it also addresses #8 and #128 (much to Miren's annoyance).

NB I have not tried to address flattening out the Tensor array here, but once this is done I think it should be relatively simple to implement.

Some questions/comments I had for discussion:

1. Do we need to have a ConstCCZ4Vars and a CCZ4Vars class? I couldn't work out how to do it but probably there is a way. 

2. Is the method of storing the vars optimal? I end up creating a Tensor object, and then storing it, whereas I think I should be able to update the value directly. For some reason this didn't work. On the other hand, what I have now is very readable.

3. We now have () brackets for the vars grid object access and [] brackets for Tensor objects like the derivatives. I am kind of ok with this because they are conceptually different and maybe this makes a useful distinction. On the other hand if they were all one or the other there might be less chance of making mistakes.

4. There is a warning about `const amrex::CellData<amrex::Real> &cell_data;` references as data members that I don't know whether to suppress.

5. The regression test fails but the differences are small, should we worry?

6. Should we remove the option of BSSN?


julianakwan:
Just reporting on some profiling that I did with the PVCs and AMD MI250Xs. For this test, I used the BinaryBH example with `params_profile.txt` and 1 MPI rank to 1 tile/GCD. Using `TinyProfile`, I found that `BinaryBHLevel::specificEvalRHS()` was about 3x slower on the PVCs on the `enhancement/refactor_vars2` branch than the `develop` branch. However, I found that on the MI250Xs, the refactored branch was about 10% faster!  Previously I had seen a 25% speed-up on both systems with the same setup using the `enhancement/vars_removal` branch. (These are all with exclusive nodes by the way, although I only used one tile/GCD in each test).

I did some more profiling with VTune, specifically Miren wanted to see the number of GPU cycles spent per line of code. 
Specifically I used: 
```
vtune -c gpu-hotspots -knob profiling-mode=source-analysis ...
```
I encountered some difficulties running the full test with `params_profile.txt` (I kept getting error messages saying that I had run out of GPU registers?) so I dropped `max_levels` to 6. 

We know from `TinyProfile` that the RHS is the most significant kernel in terms of compute time. In VTune, I can zoom in on parts of the RHS that the profiler has ranked in terms of GPU execution time. I have pasted some screen shots with the hottest lines highlighted i.e. the lines where the GPU is spending the most time within the RHS calculation, for `CCZ4RHS.impl.hpp` and `CCZ4Geometry.hpp`.  

Note that two of lines point to a left brace { and I don't know what that means. I do see some time is being spent in storing things the new way but not a significant amount, about sub-percent level. But maybe the total number of calls makes it significant? I'm not sure. 

I also ran an analysis that specifically looked for stalls: 
```
vtune -c gpu-hotspots -knob profiling-mode=source-analysis -knob source-analysis=stall-sampling ...
```
I saw what John Pennycook had reported earlier, namely that we were spending a lot of time (about 63% for `develop`) waiting for thread dependencies (called SBID stalls) to resolve. This happens when threads are waiting for a variable to updated by another thread before it can be used a calculation local to its own thread. It happens a lot for the derivatives because the stencil makes the calculations very much "non-local". Anyway, I only saw a 7% increase in the number of stalls from SBID dependencies in this branch. 

I don't find the results super helpful. I think it would be more helpful to try to understand the difference between this current approach and the previous one which seemed to be better optimized, although messier. 
Screen Shot 2025-08-26 at 10 50 46 AM
Screen Shot 2025-08-26 at 10 51 45 AM

mirenradia:
I tested the speedup of [`enhancement/refactor_vars_speedup`](https://github.com/GRTLCollaboration/GRTeclyn/tree/enhancement/refactor_vars_speedup) (actually I had to make some changes to fix CUDA compilation errors which I have done on [`test/refactor_vars_speedup_cuda_fixes`](https://github.com/GRTLCollaboration/GRTeclyn/tree/test/refactor_vars_speedup_cuda_fixes) @ 9a2f78a66cba69d76a930f2e4aad6710f4135eb0) vs `develop` at 76bb280f725428c1f0fc6cc7562f6fdd6d014bda on the new Nvidia H100 node at COSMA (`gn004` on [this page](https://cosma.readthedocs.io/en/latest/gpu.html)) using CUDA 13.0 (first time trying this out as I've been using CUDA 12.X elsewhere). I ran the BinaryBH example using `params_profile.txt` with `plot_interval = 0` and `checkpoint_interval = 0`. For each version of the code, I ran it three times and have averaged the results which are in the table below:

| Code | Time spent in RHS (s) | RHS Speed-up | Total time (s) | Total Speed-up |
| --- | --: | --: | --: | --: |
| develop @ 76bb280f725428c1f0fc6cc7562f6fdd6d014bda | 41.30 | 1.0x | 70.61 | 1.0x |
| refactor @ 9a2f78a66cba69d76a930f2e4aad6710f4135eb0 |  31.22 | 1.32x | 61.90 | 1.14x |

KAClough:
I tried to see what was causing the slowdown on GPUs. I don't think I got a lot of deep insight, but the following is a summary of what I found:

1. It is not the creation of new tensor objects that is the problem, it is really the `store_vars()` functions that create a slowdown.
2. Making the arguments a constant reference did not help.
3. Refactoring the D1 vars so it calculates all the variables and accesses their elements with getters (as for CCZ4Vars) does not make any difference to the speed, but I like the way it looks so I will keep it. It is another move away from `[]` to `()` so that is more consistent.
4. Refactoring the D2 vars so it calculates all the variables is a bad idea (presumably because it is a lot more work as most are not needed) and so we should not do it. I would like to create getters though to match the D1 format, which I hope won't break the speed up. 

My next steps are to tidy up the [enhancement/refactor_vars_speedup](https://github.com/GRTLCollaboration/GRTeclyn/tree/enhancement/refactor_vars_speedup) branch commit history, fix the matter classes and the tests, merge in @mirenradia 's CUDA fixes, and then merge that back here. At that point I think we should be ready to review it. 

The upshot of this is that the storing of vars in the `rhs` cell data has to be done visibly as `rhs[c_chi] = 1.0` which is fine for scalars but for tensors we need things like `rhs[var_idx(c_h11,i,j)] = 1.0`, which isn't as nice as `rhs.store_h(rhs_h)`, but I don't see a way around it for now, so I think we will have to accept the reduced readability, as much as it pains me to say it.

KAClough:
I can answer most of these now:

> NB I have not tried to address flattening out the Tensor array here, but once this is done I think it should be relatively simple to implement.

@julianakwan is working on this. It will mean that tensors are access with `(i,j)` not [i][j] which fits nicely with the style of the wrappers that are in this PR. 
 
> Some questions/comments I had for discussion:
> 
> 1. Do we need to have a ConstCCZ4Vars and a CCZ4Vars class? I couldn't work out how to do it but probably there is a way.

In the end we don't need this distinction as we store direct to the cell data, so all CCZ4Vars objects are const.

> 2. Is the method of storing the vars optimal? I end up creating a Tensor object, and then storing it, whereas I think I should be able to update the value directly. For some reason this didn't work. On the other hand, what I have now is very readable.

It is not optimal, see above.

> 3. We now have () brackets for the vars grid object access and [] brackets for Tensor objects like the derivatives. I am kind of ok with this because they are conceptually different and maybe this makes a useful distinction. On the other hand if they were all one or the other there might be less chance of making mistakes.

Due to the changes in the Tensor objects, this inconsistency should resolve itself.

> 4. There is a warning about `const amrex::CellData<amrex::Real> &cell_data;` references as data members that I don't know whether to suppress.

I have suppressed it. I think that we really do want a reference as a data member and not a pointer. We definitely don't want a copy as that's the whole point of the wrapper.

> 5. The regression test fails but the differences are small, should we worry?

I will check this again once we are done with the changes. 

> 6. Should we remove the option of BSSN?

Yes, or at least remove the branching as it slows things down a lot. But this is for another PR.

julianakwan:
I'm trying to benchmark the performance of this branch on Frontier and it annoying fails at start up when AMReX is trying to allocate memory to the `Arena` (although the `develop` branch is fine 🤔 ). 

Here is the output:
```
[jkwan@login10.frontier BinaryBH]$ more BinaryBH-vars-removal-3766161.out
Unloading the cpe module is insufficient to restore the system defaults.
Please run 'source
/opt/cray/pe/cpe/24.11/restore_lmod_system_defaults.[csh|sh]'. 
 
Sourcing the above script is not necessary if swapping the cpe module with a
different version. 

The following have been reloaded with a version change:
  1) gcc-native/14.2 => gcc-native/13.2     2) rocm/6.4.0 => rocm/6.2.4

The following have been reloaded with a version change:
  1) rocm/6.2.4 => rocm/6.4.0

Currently Loaded Modules:
  1) libfabric/1.22.0     5) craype/2.7.33         9) cray-mpich/8.1.31
  2) craype-network-ofi   6) cray-dsmml/0.3.0     10) craype-accel-amd-gfx90a
  3) gcc-native/13.2      7) cray-libsci/24.11.0  11) rocm/6.4.0
  4) cpe/24.11            8) PrgEnv-gnu/8.6.0

 

appending parameters: amr.plot_nfiles = -1 amr.checkpoint_nfiles = -1 checkpoint_interval = -1 plot_inte
rval=-1 amrex.use_gpu_aware_mpi=1
Initializing AMReX (25.08-19-g9d2b74990c60)...
MPI initialized with 1 MPI processes
MPI initialized with thread support level 0
Initializing HIP...
HIP initialized with 1 device.
amrex::Abort::0::HIP error in file /lustre/orion/cos004/proj-shared/jkwan/amrex/Src/Base/AMReX_Arena.cpp
 line 257 out of memory !!!
SIGABRT
MPICH ERROR [Rank 0] [job id 3766161.0] [Thu Sep 25 11:55:39 2025] [frontier03208] - Abort(6) (rank 0 in
 comm 0): application called MPI_Abort(MPI_COMM_WORLD, 6) - process 0

srun: error: frontier03208: task 0: Exited with exit code 6
srun: Terminating StepId=3766161.0
```

Here is Backtrace.0:
```
Host Name: frontier03208
=== If no file names and line numbers are shown below, one can run
            addr2line -Cpfie my_exefile my_line_address
    to convert `my_line_address` (e.g., 0x4a6b) into file name and line number.
    Or one can use amrex/Tools/Backtrace/parse_bt.py.

=== Please note that the line number reported by addr2line may not be accurate.
    One can use
            readelf -wl my_exefile | grep my_line_address'
    to find out the offset for that line.

 0: /lustre/orion/cos004/proj-shared/jkwan/GRTeclyn/Examples/BinaryBH/././main3d.hip.x86-trento.TPROF.MPI.HIP.ex() [0x45c226]
    amrex::BLBackTrace::print_backtrace_info(_IO_FILE*) at /lustre/orion/cos004/proj-shared/jkwan/amrex/Src/Base/AMReX_BLBackTrace.cpp:205

 1: /lustre/orion/cos004/proj-shared/jkwan/GRTeclyn/Examples/BinaryBH/././main3d.hip.x86-trento.TPROF.MPI.HIP.ex() [0x45be33]
    amrex::BLBackTrace::handler(int) at /lustre/orion/cos004/proj-shared/jkwan/amrex/Src/Base/AMReX_BLBackTrace.cpp:103

 2: /lustre/orion/cos004/proj-shared/jkwan/GRTeclyn/Examples/BinaryBH/././main3d.hip.x86-trento.TPROF.MPI.HIP.ex() [0x365241]
    std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_data() const at /usr/lib64/gcc/x86_64-suse-linux/13/../../../../include/c++/13/bits/basic_string.h:223
 (inlined by) std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_is_local() const at /usr/lib64/gcc/x86_64-suse-linux/13/../../../../include/c++/13/bits/basic_string.h:264
 (inlined by) std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_dispose() at /usr/lib64/gcc/x86_64-suse-linux/13/../../../../include/c++/13/bits/basic_string.h:282
 (inlined by) std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::~basic_string() at /usr/lib64/gcc/x86_64-suse-linux/13/../../../../include/c++/13/bits/basic_string.h:804
 (inlined by) amrex::Arena::allocate_system(unsigned long) at /lustre/orion/cos004/proj-shared/jkwan/amrex/Src/Base/AMReX_Arena.cpp:?

 3: /lustre/orion/cos004/proj-shared/jkwan/GRTeclyn/Examples/BinaryBH/././main3d.hip.x86-trento.TPROF.MPI.HIP.ex() [0x36a82a]
    amrex::CArena::alloc_protected(unsigned long) at /lustre/orion/cos004/proj-shared/jkwan/amrex/Src/Base/AMReX_CArena.cpp:?

 4: /lustre/orion/cos004/proj-shared/jkwan/GRTeclyn/Examples/BinaryBH/././main3d.hip.x86-trento.TPROF.MPI.HIP.ex() [0x36a599]
    amrex::CArena::alloc(unsigned long) at /lustre/orion/cos004/proj-shared/jkwan/amrex/Src/Base/AMReX_CArena.cpp:?

 5: /lustre/orion/cos004/proj-shared/jkwan/GRTeclyn/Examples/BinaryBH/././main3d.hip.x86-trento.TPROF.MPI.HIP.ex() [0x366d51]
    amrex::Arena::Initialize(bool) at /lustre/orion/cos004/proj-shared/jkwan/amrex/Src/Base/AMReX_Arena.cpp:409

 6: /lustre/orion/cos004/proj-shared/jkwan/GRTeclyn/Examples/BinaryBH/././main3d.hip.x86-trento.TPROF.MPI.HIP.ex() [0x2c0ff7]
    amrex::Initialize(int&, char**&, bool, int, std::function<void ()> const&, std::ostream&, std::ostream&, void (*)(char const*)) at /lustre/orion/cos004/proj-shared/jkwan/amrex/Src/Base/AMReX.cpp:678

 7: /lustre/orion/cos004/proj-shared/jkwan/GRTeclyn/Examples/BinaryBH/././main3d.hip.x86-trento.TPROF.MPI.HIP.ex() [0x5cf4a6]
    mainSetup(int, char**) at /lustre/orion/cos004/proj-shared/jkwan/GRTeclyn/Source/GRTeclynCore/SetupFunctions.hpp:40 (discriminator 8)

 8: /lustre/orion/cos004/proj-shared/jkwan/GRTeclyn/Examples/BinaryBH/././main3d.hip.x86-trento.TPROF.MPI.HIP.ex() [0x5cf92d]
    main at /lustre/orion/cos004/proj-shared/jkwan/GRTeclyn/Examples/BinaryBH/Main_BinaryBH.cpp:81

 9: /lib64/libc.so.6(+0x40e6c) [0x7fffcfcbce6c]

10: /lib64/libc.so.6(__libc_start_main+0x87) [0x7fffcfcbcf35]

11: /lustre/orion/cos004/proj-shared/jkwan/GRTeclyn/Examples/BinaryBH/././main3d.hip.x86-trento.TPROF.MPI.HIP.ex() [0x2bf961]
    _start
../sysdeps/x86_64/start.S:117


===== TinyProfilers ======
```
Not very helpful is it?

mirenradia:
@julianakwan, note that @KAClough has [set `amrex.the_arena_init_size` explicitly in `params_profile.txt`](https://github.com/GRTLCollaboration/GRTeclyn/pull/138/files#diff-d6366773ec57f5effc0919b614ad57b7471d9d94c402e3ec26cbc28d5f257e53R11). I'm not sure why this was committed but this is probably the cause of your out of memory error. You should remove it and let AMReX do its default thing (i.e. allocate 3/4 of the GPU memory).

julianakwan:
> @julianakwan, note that @KAClough has [set `amrex.the_arena_init_size` explicitly in `params_profile.txt`](https://github.com/GRTLCollaboration/GRTeclyn/pull/138/files#diff-d6366773ec57f5effc0919b614ad57b7471d9d94c402e3ec26cbc28d5f257e53R11). I'm not sure why this was committed but this is probably the cause of your out of memory error. You should remove it and let AMReX do its default thing (i.e. allocate 3/4 of the GPU memory).

Yes I can confirm that this was the problem! There should really be a :face-palm: reaction as I should have checked `params_profile.txt` more carefully!

julianakwan:
I have some benchmarks to report. This was done on Frontier with 1 MPI rank to 1 GCD and each result has been averaged over 3 runs:

| Code | Time spent in RHS (s) | RHS Speed-up | Total time (s) | Total Speed-up |
| --- | --: | --: | --: | --: |
| develop @ 76bb280 | 157.83 | 1.0x | 198.46 | 1.0x |
| refactor_vars2 @ 72ffeba | 112.73 | 1.4x | 155.066 | 1.28x |

This is comparable to the performance that I had measured previously on Frontier with the `enhancement/vars_removal` branch

mirenradia:
> This is comparable to the performance that I had measured previously on Frontier with the `enhancement/refactor_vars` branch

Do you mean the [`enhancement/vars_removal`](https://github.com/GRTLCollaboration/GRTeclyn/tree/enhancement/vars_removal) branch? I think the `enhancement/refactor_vars` branch is an older version of the stuff on this branch that Katy worked on.

KAClough:
> > @julianakwan, note that @KAClough has [set `amrex.the_arena_init_size` explicitly in `params_profile.txt`](https://github.com/GRTLCollaboration/GRTeclyn/pull/138/files#diff-d6366773ec57f5effc0919b614ad57b7471d9d94c402e3ec26cbc28d5f257e53R11). I'm not sure why this was committed but this is probably the cause of your out of memory error. You should remove it and let AMReX do its default thing (i.e. allocate 3/4 of the GPU memory).
> 
> Yes I can confirm that this was the problem! There should really be a :face-palm: reaction as I should have checked `params_profile.txt` more carefully!

I'm sorry for this, it is me who shouldn't have committed it! It is a setting I have to use on the Cosma AMD GPU for reasons I have not understood. I will remove it when I next update the branch.

julianakwan:
> > This is comparable to the performance that I had measured previously on Frontier with the `enhancement/refactor_vars` branch
> 
> Do you mean the [`enhancement/vars_removal`](https://github.com/GRTLCollaboration/GRTeclyn/tree/enhancement/vars_removal) branch? I think the `enhancement/refactor_vars` branch is an older version of the stuff on this branch that Katy worked on.

Yes, I will correct the post.

julianakwan:
> > > @julianakwan, note that @KAClough has [set `amrex.the_arena_init_size` explicitly in `params_profile.txt`](https://github.com/GRTLCollaboration/GRTeclyn/pull/138/files#diff-d6366773ec57f5effc0919b614ad57b7471d9d94c402e3ec26cbc28d5f257e53R11). I'm not sure why this was committed but this is probably the cause of your out of memory error. You should remove it and let AMReX do its default thing (i.e. allocate 3/4 of the GPU memory).
> > 
> > 
> > Yes I can confirm that this was the problem! There should really be a :face-palm: reaction as I should have checked `params_profile.txt` more carefully!
> 
> I'm sorry for this, it is me who shouldn't have committed it! It is a setting I have to use on the Cosma AMD GPU for reasons I have not understood. I will remove it when I next update the branch.

No problem! As I understand it, the Cosma8 MI300Xs are on a shared node in some kind of free-for-all Wild West configuration, so you probably had to manually request enough memory.

julianakwan:
I have some more benchmarks to report (yes I know this should be its own issue). Same set up as above on Frontier:

| Code | Time spent in RHS (s) | RHS Speed-up | Total time (s) | Total Speed-up |
| --- | --: | --: | --: | --: |
| develop @ 76bb280 | 157.83 | 1.0x | 198.46 | 1.0x |
| refactor_tensors @ ce284cb | 155.56 | 1.01x | 196.2 | 1.01x |

😞

mirenradia:
I have some bad news to report.

I tested this branch @ 72ffeba1bbe7097227e33d913c06846c34cc81dc vs `develop` at 76bb280f725428c1f0fc6cc7562f6fdd6d014bda on a single GPU (Intel Data Center GPU Max 1550/PVC) on Dawn (2 MPI ranks so there is 1 per stack). I ran the BinaryBH example using `params_profile.txt` with `plot_interval = 0` and `checkpoint_interval = 0`. For each version of the code, I ran it three times and have averaged the results which are in the table below:

| Code | Time spent in RHS (s) | RHS Speed-up | Total time (s) | Total Speed-up |
| --- | --: | --: | --: | --: |
| develop @ 76bb280f725428c1f0fc6cc7562f6fdd6d014bda | 50.16 | 1.0x | 77.89 | 1.0x |
| refactor @ 72ffeba1bbe7097227e33d913c06846c34cc81dc | 124.93 | 0.40x | 150.5 | 0.52x |

I think I would accept a small degradation in performance on PVCs since GRTeclyn is more likely to be run on Nvidia and AMD GPUs but I think this is probably too big to accept. I will try and investigate.

Examples/BinaryBH/BinaryBHLevel.cpp
```diff
@@ -304,7 +315,7 @@ void BinaryBHLevel::specificPostTimeStep()
         // writeout_level
         bool write_punctures = at_level_timestep_multiple(
             simParams().puncture_tracking_writeout_level);
-        amrex::Real cur_time = get_state_data(State_Type).curTime();
+        amrex::Real cur_time = get_state_data(state_index).curTime();
```
julianakwan: Could you please change `cur_time` to `current_time` to be consistent with L212? Thanks!

Examples/BinaryBH/params_profile.txt
```diff
@@ -8,14 +8,20 @@
 
 verbosity = 1
 
+amrex.the_arena_init_size = 190000000000
```
julianakwan: I think this should be removed because it will segfault on GPUs that don't have that much memory. (I think this was peculiar to the MI300X's).

Examples/BinaryBH/params_profile.txt
```diff
@@ -80,7 +86,7 @@ nonzero_asymptotic_values = 1.0 1.0 1.0 1.0 1.0
 # dt will be dx*dt_multiplier on each grid level
 dt_multiplier = 0.25
 # stop_time = 2.0
-max_steps = 1
+max_steps = 10
```
julianakwan: I think you can leave it on 1 for profiling. 10 seems a bit much.

Source/CCZ4/CCZ4D2Vars.hpp
```diff
@@ -0,0 +1,76 @@
...
+        calculate_d2_derivs(ix, iy, iz, state, a_deriv);
+    }
+
+    // default empty contructor - for tests
+    AMREX_GPU_HOST_DEVICE CCZ4D2Vars()
+    {
+        FOR (k, l)
+        {
+            chi[k][l] = 0.0;
+            // Theta[k][l] = 0.0;
```
julianakwan: Could you please remove the ones that aren't needed? Thanks!

Source/CCZ4/CCZ4Geometry.hpp
```diff
@@ -26,172 +28,345 @@ struct ricci_t
...
+    {
+        Aij_squared += inverse_metric[i][k] * inverse_metric[j][l] *
+                       vars.A(i, j) * vars.A(k, l);
+    }
+    return Aij_squared;
+}
+
+// This is A_ij A^ij
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE amrex::Real
+compute_Aij_squared(const CCZ4Vars &vars,
```
julianakwan, line 126: I noticed that in `CCZ4RHS` we don't actually need the to compute the inverse metric because we already calculate `A_UU`. You might want to add overload `compute_Aij_squared` so that `A_UU` can be passed in.

Source/CCZ4/CCZ4StateVariables.hpp
```diff
@@ -52,6 +52,14 @@ enum
     NUM_CCZ4_VARS
 };
 
+// A function to return the right index for the tensors based on the
+// ordering below 0: T11, 1: T12, 2: T13, 3: T22, 4: T23, 5: T33
+[[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE int var_idx(int ivar,
```
julianakwan: I think this could go somewhere else as a macro...because I find myself using it for any symmetric data structure, e.g. the derivatives, as well. I'm not entirely sure where though 😅 
Also, sometimes, I'll pass in ivar=0 when I don't need the CCZ4 variables, maybe that can be a defaulted variable, like `int var_idx(int i, int j, int ivar=0)`.

Source/CCZ4/Weyl4.hpp
```diff
@@ -59,73 +60,66 @@ class Weyl4
...
     //! Constructor of class Weyl4
     /*!
         Takes in the centre for the calculation of the tetrads, grid spacing and
         the formulation.
     */
+    // TODO: Remove dependence on formulation?
     Weyl4(const std::array<double, AMREX_SPACEDIM> &a_center, double a_dx,
-          int a_dcomp, int a_formulation = CCZ4RHS<>::USE_CCZ4)
-        : m_center(a_center), m_dx(a_dx), m_deriv(a_dx), m_dcomp(a_dcomp),
+          int a_out_comp, int a_formulation = CCZ4RHS<>::USE_CCZ4)
```
julianakwan, line 74: Sorry to be a pain...but why is the `dcomp` -> `comp` change necessary? 😅 I think we should leave it as `dcomp` for consistency with AMReX and other AMReX codes.

Source/Maths/TensorAlgebra.hpp
```diff
@@ -325,6 +325,37 @@ compute_christoffel(const Tensor<2, Tensor<1, amrex::Real>> &d1_metric,
     return out;
 }
 
+/// Computes the conformal christoffel symbol
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE chris_t compute_christoffel(
```
julianakwan, line 329: Is it necessary to keep the old definition? I just can't see us using it anywhere...

Source/Matter/ConstraintsWithMatter.hpp
```diff
@@ -56,8 +39,9 @@ template <class matter_t> class ConstraintsWithMatter : public Constraints
     //! The compute member which calculates the constraints at each point in the
     //! box
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
-    compute(int i, int j, int k, const amrex::Array4<amrex::Real> &out_arrays,
-            const amrex::Array4<amrex::Real const> &state_arrays) const;
+    operator()(int ix, int iy, int iz,
+               const amrex::Array4<amrex::Real> &constraints,
+               const amrex::Array4<amrex::Real const> &state) const;
```
julianakwan, line 44: Shouldn't it be `state_arrays` for consistency with `BinaryBHLevel`? 😅

Source/Matter/EMTensor.hpp
```diff
@@ -57,7 +56,7 @@ template <class matter_t, enum EMTensorOptions em_tensor_options> class EMTensor
 
     AMREX_FORCE_INLINE static void
     compute_mf(amrex::MultiFab &out_mf, int dcomp, int ncomp,
```
julianakwan, line 58: Should this `dcomp` be `out_comp` for consistency with the others?

Source/Grids/FourthOrderDerivatives.hpp
```diff
@@ -28,10 +26,10 @@ class FourthOrderDerivatives
     }
 
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE amrex::Real
-    diff1(const amrex::Real *in_ptr, const int idx, const int stride) const
+    diff1(const amrex::Real *in_ptr, const int stride, const int idx = 0) const
     {
-        amrex::Real weight_far  = 8.33333333333333333333e-2_rt;
-        amrex::Real weight_near = 6.66666666666666666667e-1_rt;
+        amrex::Real weight_far  = 8.33333333333333333333e-2;
```
julianakwan: I'm confused as to why these are no longer AMReX literals?

Source/Grids/FourthOrderDerivatives.hpp
```diff
@@ -40,35 +38,90 @@ class FourthOrderDerivatives
...
+        const auto *var_ptr = state_ptr_xyz + ivar * state.nstride;
+        FOR (idir)
+        {
+            d1[idir] = diff1(var_ptr, strides[idir]);
+        }
         return d1;
     }
 
+    [[nodiscard]] AMREX_GPU_DEVICE AMREX_FORCE_INLINE Tensor<1, amrex::Real>
+    diff1_scalar(int ix, int iy, int iz,
```
julianakwan, line 68: I think `diff1_scalar`, `diff1_vector` and `diff1_tensor` could probably be templated over the Tensor rank. On the other hand, I do make some significant changes to how the derivatives are calculated in the `enhancement/refactor_vars2_kernel_fission` branch so I'm not sure if this is something that we should worry about in this PR. 

Same for the `diff2` and `advection` calculations.

Source/CCZ4/Constraints.impl.hpp
```diff
@@ -128,8 +199,9 @@ void Constraints::set_up(int a_state_index, bool a_calc_mom_norm)
...
         &amrex::cell_quartic_interp);
 
-    // We only need the non-gauge CCZ4 variables to calculate the constraints
-    derive_lst.addComponent(name, desc_lst, a_state_index, 0, c_lapse);
+    // Get all the vars to allow us to use the CCZ4Vars class
+    // Technically not all needed but probably doesn't hurt performance
+    derive_lst.addComponent(name, desc_lst, a_state_index, 0, NUM_VARS);
 }
 
 void Constraints::compute_mf(amrex::MultiFab &out_mf, int dcomp, int ncomp,
```
julianakwan, line 207: I think `dcomp` has been renamed to something else in this PR.

Source/CCZ4/Constraints.impl.hpp
```diff
@@ -128,8 +199,9 @@ void Constraints::set_up(int a_state_index, bool a_calc_mom_norm)
         [=](const amrex::Box &box) { return amrex::grow(box, num_ghosts); },
         &amrex::cell_quartic_interp);
 
-    // We only need the non-gauge CCZ4 variables to calculate the constraints
-    derive_lst.addComponent(name, desc_lst, a_state_index, 0, c_lapse);
+    // Get all the vars to allow us to use the CCZ4Vars class
+    // Technically not all needed but probably doesn't hurt performance
+    derive_lst.addComponent(name, desc_lst, a_state_index, 0, NUM_VARS);
```
julianakwan: Why do we now pass in all the variables? I agree that it probably doesn't hurt performance by much, but I think AMReX allocates an intermediary `MultiFab` of that size. 

I think it's also not great practice from a software "tidiness" perspective because the derived variable should only have access to the state variables that it needs.

Source/CCZ4/Weyl4.impl.hpp
```diff
@@ -11,49 +11,53 @@
...
+    weyl_scalars(ix, iy, iz, m_out_comp + 1) = out.Im;
 }
 
 AMREX_GPU_DEVICE AMREX_FORCE_INLINE Tensor<3, amrex::Real>
-Weyl4::compute_epsilon3_LUU(const Vars<amrex::Real> &vars,
+Weyl4::compute_epsilon3_LUU(const CCZ4Vars &vars,
                             const Tensor<2, amrex::Real> &h_UU) const
 {
     // raised normal vector, NB index 3 is time
     Tensor<1, amrex::Real, 4> n_U;
```
julianakwan: Could you please replace the "4" with `GR_SPACEDIM`+1? I'm trying to reduce the usage of magic numbers :)

Source/CCZ4/Weyl4.impl.hpp
```diff
@@ -75,7 +79,8 @@ Weyl4::compute_epsilon3_LUU(const Vars<amrex::Real> &vars,
         for (int l = 0; l < 4; ++l)
```
julianakwan: Same here :)

Source/CCZ4/Weyl4.impl.hpp
```diff
@@ -295,8 +298,8 @@ AMREX_GPU_DEVICE AMREX_FORCE_INLINE Tetrad_t Weyl4::compute_null_tetrad(
     amrex::Real omega_23 = 0.0;
     FOR (i, j)
     {
-        omega_13 += out.v[i] * out.w[j] * vars.h[i][j] / vars.chi;
-        omega_23 += out.u[i] * out.w[j] * vars.h[i][j] / vars.chi;
+        omega_13 += out.v[i] * out.w[j] * vars.h(i, j) / vars.chi();
```
julianakwan, line 301: I think some of these loops can be combined together (while remembering that `omega_23` needs to be divided by `1/sqrt(omega_22)` 😅 ) (Also I think this should be `std::sqrt` - there might be potential for confusion when using SYCL, I don't remember if it defaults to `sycl::sqrt` or `std::sqrt` when the namespace isn't specified so best to specify to be sure).

Source/Matter/ScalarFieldAdvecVars.hpp
```diff
@@ -0,0 +1,50 @@
...
+{
+  public:
+    // NOLINTBEGIN(cppcoreguidelines-pro-type-member-init)
+    AMREX_GPU_DEVICE
+    ScalarFieldAdvecVars(int ix, int iy, int iz,
+                         const amrex::Array4<const amrex::Real> &state,
+                         const FourthOrderDerivatives &a_deriv)
+        : CCZ4AdvecVars(ix, iy, iz, state, a_deriv)
+    {
+        Tensor<1, amrex::Real> shift_vector;
```
julianakwan: If you make the shift vector a member variable of `CCZ4AdvecVars` I think you could remove this calculation and reduce code duplication ...? I'm also not sure that `m_advec_state` should be calculated for all `ScalarField` vars or on a as needed basis...

Source/Matter/Weyl4WithMatter.hpp
```diff
@@ -34,14 +31,14 @@ template <class matter_t> class Weyl4WithMatter : public Weyl4
     //! The compute member which calculates the wave quantities at each point on
     //! the grid
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
-    compute(int i, int j, int k,
-            const amrex::Array4<amrex::Real> &derive_arrays,
-            const amrex::Array4<amrex::Real const> &state_arrays) const;
+    operator()(int ix, int iy, int iz,
+               const amrex::Array4<amrex::Real> &weyl_scalars,
+               const amrex::Array4<amrex::Real const> &state) const;
```
julianakwan, line 36: `state` or `state_arrays`?

Tests/BSSNMatterTest/BSSNMatterTest.cpp
```diff
@@ -129,9 +129,8 @@ void run_bssn_matter_test()
         amrex::ParallelFor(out_mf,
                            [=] AMREX_GPU_DEVICE(int ibox, int i, int j, int k)
```
julianakwan: These indices could be `ix`, `iy` and `iz` also.

Tests/CCZ4GeometryUnitTests/CCZ4GeometryMathematicaValues.hpp
```diff
@@ -4,200 +4,176 @@
...
+d2.h[2][2][0][0]                       = 0.29285200380140775;
+d2.h[2][2][1][0]                       = 0.12090991364315329;
+d2.h[2][2][2][0]                       = 0.11839822831093305;
+d2.h[2][2][0][1]                       = 0.12090991364315329;
+d2.h[2][2][1][1]                       = 0.5777485076926592;
+d2.h[2][2][2][1]                       = 0.24842636089202075;
+d2.h[2][2][0][2]                       = 0.11839822831093305;
+d2.h[2][2][1][2]                       = 0.24842636089202075;
+d2.h[2][2][2][2]                       = 0.2540349225617704;
+// d2.Gamma[0][0][0]       = 0.5832629795220816;
```
julianakwan: If these aren't needed, we should remove them.

Tests/CCZ4GeometryUnitTests/CCZ4GeometryUnitTest.cpp
```diff
@@ -5,74 +5,161 @@
...
+        const auto &geometry_array                 = geometry_fab.array();
+
+        amrex::ParallelFor(box,
+                           [=] AMREX_GPU_DEVICE(int i, int j, int k)
+                           {
+                               const amrex::IntVect iv{i, j, k};
+                               compute_ccz4_test_geometry(in_array, iv,
+                                                          geometry_array);
+                           });
+
```
julianakwan, line 115: Could you please add a `Gpu::streamSynchronize` after the `amrex::ParallelFor`? The GPU threads only sync up after a `MFIter` loop and not after any other parallel construct. I think this is why the CI on the A100s are failing. At least, I added it in and it passed on Tursa and Polaris. 😅 

Also please change the indices to `ix, iy, iz`. Thanks!

Tests/CCZ4GeometryUnitTests/CCZ4GeometryUnitTest.hpp
```diff
@@ -5,6 +5,12 @@
 #ifndef CCZ4GEOMETRYUNITTEST_HPP_
 #define CCZ4GEOMETRYUNITTEST_HPP_
 
+#include "DimensionDefinitions.hpp"
+
+constexpr int NUM_GEOMETRY_TEST_VARS = GR_SPACEDIM * GR_SPACEDIM * GR_SPACEDIM +
```
julianakwan, line 12: I don't know what this is supposed to be ...? 🤣 Maybe a comment would be nice

Tests/CCZ4RHSTest/CCZ4RHSTest.cpp
```diff
@@ -106,17 +104,18 @@ void run_ccz4_rhs_test()
...
         amrex::ParallelFor(
             box,
-            [=] AMREX_GPU_DEVICE(int i, int j, int k)
+            [=] AMREX_GPU_DEVICE(int ix, int iy, int iz)
             {
-                current_ccz4_rhs.compute(i, j, k, current_out_array,
-                                         in_c_array);
-                old_ccz4_rhs.compute(i, j, k, old_out_array, in_c_array);
+                old_ccz4_rhs.compute(ix, iy, iz, old_out_array, in_c_array);
+                // std::cout << "testing here bug start\n";
```
julianakwan: Could you remove these test comments please?

Tests/CCZ4RHSTest/CCZ4RHSTest.hpp
```diff
@@ -5,6 +5,8 @@
 #ifndef CCZ4RHSTEST_HPP_
 #define CCZ4RHSTEST_HPP_
 
+#include <AMReX_Print.H>
```
julianakwan: If you're not printing test statements, you can take this out.

julianakwan (changes requested):
Thanks very much for this massive PR! 

I only had a few minor changes to suggest. Most importantly: 

- I think there's a lot of code in the wrappers that could be expressed more concisely. Perhaps the three wrappers can be combined into a single one with templating? I'm not sure exactly how this can be achieved, but I'm happy to take a look at it if you like. 
- The remaining CI issue can be resolved by adding `amrex::Gpu::streamSynchronize()` immediately after the call  to `ParallelFor`. 

NB: I still haven't had a chance to test the performance on Intel GPUs for this latest iteration. Unfortunately Aurora is down for maintenance this week, so I will do it while you are address the changes in the current version.

KAClough:
> Thanks very much for this massive PR!

No thank you for reviewing it! I will comment on the individual comments and then start to update them.

> I only had a few minor changes to suggest. Most importantly:
> 
> * I think there's a lot of code in the wrappers that could be expressed more concisely. Perhaps the three wrappers can be combined into a single one with templating? I'm not sure exactly how this can be achieved, but I'm happy to take a look at it if you like.

Hmm, I do understand this, but I'm not sure that I like combining them, mainly because logically they aren't all the same thing - e.g. the advection wrapper is fundamentally different to CCZ4Vars wrapper, and when people extend the code they might want to use wrappers in different ways. If we introduce templating then it does restrict what you can do. I would advocate leaving this for now and focussing on resolving the other issues, especially the ones which affect the performance. 

> * The remaining CI issue can be resolved by adding `amrex::Gpu::streamSynchronize()` immediately after the call  to `ParallelFor`.

Thank you for spotting this!

> NB: I still haven't had a chance to test the performance on Intel GPUs for this latest iteration. Unfortunately Aurora is down for maintenance this week, so I will do it while you are address the changes in the current version.

Sounds good.

mirenradia:
I haven't had time to look at this PR yet but I agree with @julianakwan that there's a lot of duplication in all of the wrapper classes. Perhaps not all of them can be combined but the `CCZ4AdvecVars` and the `CCZ4D1Vars` classes look very similar.

Source/CCZ4/CCZ4StateVariables.hpp
KAClough: Yeah I agree, it doesn't 100% belong here, maybe it should be its own small file that just gets included here? Or shall we make it a macro and add it to [`DimensionDefinitions.hpp`](https://github.com/GRTLCollaboration/GRTeclyn/blob/enhancement/refactor_vars2/Source/Grids/DimensionDefinitions.hpp)? The `FOR2_SYM` macro kind of assumes this, so maybe it makes sense to make them in the same place.

Agree about default `var=0`.

Source/CCZ4/Weyl4.hpp
KAClough: What does the d in `dcomp` stand for? I think I changed this to be more consistent with GRChombo naming, and because I didn't know what the d was for. But I don't feel strongly as long as we are consistent throughout.

Source/Maths/TensorAlgebra.hpp
KAClough: Hmm, it is a general tensor algebra function that could be applied to any metric without knowing we are in CCZ4 - we were using it in the code that sets Gamma^i in the Kerr example (but I guess that might change), and I think we use it in GRDzhaDzha, which doesn't store the metric so would need it as a general tensor function. I will remove the computation of the physical christoffel symbol `compute_phys_chris` here as that assumes CCZ4, but I'd be inclined to keep the general christoffel calculation as it is likely to be useful.

Source/Matter/ConstraintsWithMatter.hpp
KAClough: I've completely forgotten the convention we decided on, I'll check but you are certainly right! Whatever we pick is ok as long as it is consistent.

Source/Matter/EMTensor.hpp
KAClough: I guess it depends if we want to keep dcomp everywhere, but yes I like `out_comp` better.

Source/Grids/FourthOrderDerivatives.hpp
KAClough: I think this is just a mistake, probably from rebasing nightmares...

Source/Grids/FourthOrderDerivatives.hpp
KAClough: Yes, I think let's leave that for you to do in your PR, it probably makes more sense if you are changing the derivs anyway.

Source/CCZ4/Constraints.impl.hpp
KAClough: As above

Source/Matter/Weyl4WithMatter.hpp
KAClough: As above.

Source/Matter/ScalarFieldAdvecVars.hpp
KAClough: Ok, good idea for the shift. I think we do always want the advection for all variables, as it is always added in each RHS.

Examples/BinaryBH/BinaryBHLevel.cpp
KAClough: Done

Examples/BinaryBH/params_profile.txt
KAClough: Done

Examples/BinaryBH/params_profile.txt
KAClough: Done

Source/CCZ4/CCZ4D2Vars.hpp
KAClough: Done

Source/CCZ4/Constraints.impl.hpp
KAClough: The issue here is that we want to use the CCZ4Vars wrapper for the expressions, so I think you need all the variables to be passed. We could also create a ConstraintsVars wrapper with fewer variables, but then that would be another wrapper class...

Source/Matter/ScalarFieldAdvecVars.hpp
KAClough: I've updated this but please check what I did. In particular, I was wondering if we are calculating the advection components for all variables twice because we call the constructor for the CCZ4AdvecVars, and then call the constructor here?

Source/CCZ4/CCZ4Geometry.hpp
KAClough: Hmm, but the problem with the overload is that it isn't then clear if you are passing in the inverse metric or A_UU as the second argument since both are rank 2 tensors... I would prefer to pass in the inverse metric as that's more general. I thought that for GPUs calculations were cheap so it didn't matter too much?

Source/Matter/ConstraintsWithMatter.hpp
KAClough: I think I have consistently used `state` in all the classes that have an `operator()` function called by the parallel4 loops, and `state_arrays` when I pass them into the parallel4 construct in the Level class (because there we have to differentiate between the state and the arrays constructed from them). I agree it is not completely consistent but I think adding the word array everywhere at this level just makes the naming longer and doesn't really add much information as here you can see it is an array. So I'd be happy to leave it this way.

Source/Grids/FourthOrderDerivatives.hpp
KAClough: Done

Source/CCZ4/Weyl4.impl.hpp
KAClough: Actually I don't think it makes sense to do this because the code is otherwise hardcoded for 3+1 dimensions (e.g. we use the 3d and 4d epsilons etc). So people should not be using this class with a different number of dimensions (or we need to fully rewrite the code to do GWs in a different number of dimensions, but this is non trivial).

Source/CCZ4/Weyl4.impl.hpp
KAClough: As above

Source/CCZ4/Weyl4.impl.hpp
KAClough: I've updated the `sqrt` but I don't like the idea of combining the loops as it would reduce readability of the GSO method. The way it is coded is the textbook way, and if it doesn't degrade performance a lot (which it shouldn't as this is a derived variable anyway) I'd rather leave it as it is.

Tests/BSSNMatterTest/BSSNMatterTest.cpp
KAClough: Done

Tests/CCZ4GeometryUnitTests/CCZ4GeometryMathematicaValues.hpp
KAClough: Done

Tests/CCZ4GeometryUnitTests/CCZ4GeometryUnitTest.cpp
KAClough: Done, I will push and hope for the best!

Tests/CCZ4GeometryUnitTests/CCZ4GeometryUnitTest.hpp
KAClough: Done

Tests/CCZ4RHSTest/CCZ4RHSTest.cpp
KAClough: Done

Tests/CCZ4RHSTest/CCZ4RHSTest.hpp
KAClough: For some reason if I do this is breaks the compilation...

Examples/BinaryBH/params_profile.txt
```diff
@@ -8,14 +8,20 @@
 
 verbosity = 1
 
+# Uncomment if using GPUs
+# amrex.use_gpu_aware_mpi = 1
+
```
mirenradia: AMReX added support for parameters that are conditional on `AMREX_USE_GPU` in AMReX-Codes/amrex#4927 so we can just change this to
```suggestion
#ifdef AMREX_USE_GPU
amrex.use_gpu_aware_mpi = 1
#endif
```

Source/CCZ4/CCZ4D1Vars.hpp
```diff
@@ -0,0 +1,113 @@
...
+        {
+            FOR (i)
+            {
+                m_d1_state[ivar][i] = 0.0;
+            }
+        }
+    }
+    // NOLINTEND(cppcoreguidelines-pro-type-member-init)
+
+    amrex::GpuArray<Tensor<1, amrex::Real>, NUM_VARS> m_d1_state;
```
mirenradia: Why do we calculate and store `NUM_VARS` variables rather than `NUM_CCZ4_VARS`? Obviously in the vacuum case, it doesn't make a difference but in the matter case, it matters more (pun intended).

Source/CCZ4/CCZ4AdvecVars.hpp
```diff
@@ -0,0 +1,91 @@
...
+        {
+            m_shift_vector[idir] = state(ix, iy, iz, c_shift1 + idir);
+        }
+
+        // Calculate the advec quantities for all vars
+        m_advec_state = a_deriv.advec_state(ix, iy, iz, state, m_shift_vector);
+    }
+    // NOLINTEND(cppcoreguidelines-pro-type-member-init)
+
+    amrex::GpuArray<amrex::Real, NUM_VARS> m_advec_state;
```
mirenradia: (Same comment as CCZ4D1Vars)
Why do we calculate and store `NUM_VARS` variables rather than `NUM_CCZ4_VARS`? Obviously in the vacuum case, it doesn't make a difference but in the matter case, it matters more (pun intended).

Source/CCZ4/CCZ4Geometry.hpp
KAClough: Give it a different name.

Source/CCZ4/Weyl4.hpp
KAClough: d=destination

Source/CCZ4/Constraints.impl.hpp
KAClough: NUM_CCZ4_VARS or just c_lapse + comment

Source/CCZ4/Weyl4.impl.hpp
KAClough: GR_SPACEDIM static assert = 3

Source/Matter/ScalarFieldAdvecVars.hpp
KAClough: Do them separately

Source/CCZ4/CCZ4D1Vars.hpp
KAClough: Leave for Juliana updates

Source/CCZ4/CCZ4AdvecVars.hpp
KAClough: As above

Source/CCZ4/CCZ4D1Vars.hpp
mirenradia: Having spoken to @julianakwan, I don't think she plans to change this file so can this be changed in this PR instead of waiting for hers.

Source/CCZ4/CCZ4AdvecVars.hpp
mirenradia: (Same comment as CCZ4D1Vars)

mirenradia:
Was there a reason we don't use a wrapper for the RHS in `CCZ4RHS`? I don't want to introduce another wrapper but maybe it's possible to modify the existing `CCZ4Vars` wrapper to handle both the const and non-const cases.

KAClough:
> Was there a reason we don't use a wrapper for the RHS in `CCZ4RHS`? I don't want to introduce another wrapper but maybe it's possible to modify the existing `CCZ4Vars` wrapper to handle both the const and non-const cases.

Yes, because when we did it introduced a big slowdown for when we tried to store the vars. Maybe @SamuelBrady has more to add as he also looked at this.

Examples/BinaryBH/params_profile.txt
KAClough: Done

Examples/BinaryBH/params_profile.txt
KAClough: Amended as suggested by Miren below.

Source/CCZ4/CCZ4Geometry.hpp
KAClough: Done

Source/CCZ4/CCZ4StateVariables.hpp
KAClough: Done

Tests/CCZ4GeometryUnitTests/CCZ4GeometryUnitTest.hpp
KAClough: Not sure why this is not showing but it has been updated.

Source/Matter/EMTensor.hpp
KAClough: I have made it `dcomp` in the `compute_mf()` function, but in the class constructor I use `out_comp`, which is where `dcomp` is passed. I think this is a good compromise as you can follow the code from the AMReX naming to the normal human naming.

KAClough: To be honest I think the naming in the derived classes is all pretty terrible, but I think we should revisit this as a separate issue.

Source/CCZ4/Constraints.impl.hpp
KAClough: Left as dcomp

Source/CCZ4/Constraints.impl.hpp
KAClough: On reflection I don't like the idea of cutting off the gauge variables if we then use the CCZ4Vars wrapper, it is just confusing. If we do something different with the wrappers maybe we can revisit this. I have just made it `NUM_CCZ4_VARS`

Source/CCZ4/Weyl4.impl.hpp
KAClough: Done

Source/CCZ4/Weyl4.impl.hpp
KAClough: Done

Source/Grids/FourthOrderDerivatives.hpp
```diff
@@ -108,134 +164,209 @@ class FourthOrderDerivatives
...
+                     const int idx = 0) const
     {
-        amrex::Real weight_vfar  = 1.56250e-2_rt;
-        amrex::Real weight_far   = 9.37500e-2_rt;
-        amrex::Real weight_near  = 2.34375e-1_rt;
-        amrex::Real weight_local = 3.12500e-1_rt;
+        amrex::Real weight_vfar  = 1.56250e-2;
+        amrex::Real weight_far   = 9.37500e-2;
+        amrex::Real weight_near  = 2.34375e-1;
+        amrex::Real weight_local = 3.12500e-1;
```
mirenradia: I think this was another unintentional removal of `_rt`? Not sure if you've fixed this one in any unpushed commits but thought I'd point it out here just in case.

Source/CCZ4/CCZ4D1Vars.hpp
KAClough: Done

Source/CCZ4/CCZ4AdvecVars.hpp
KAClough: Done

Source/Matter/ScalarFieldAdvecVars.hpp
KAClough: Done

Source/Grids/FourthOrderDerivatives.hpp
KAClough: Done.

Source/CCZ4/CCZ4RHS.hpp
```diff
@@ -88,32 +81,19 @@ class CCZ4RHS
...
     /** This function orchestrates the calculation of the rhs for one specific
-     * grid cell. This function is called by the BoxLoops::loop for each grid
-     * cell; there should rarely be a need to call it directly.
+     * grid cell.
      */
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
-    compute(int i, int j, int k, const amrex::Array4<amrex::Real> &rhs,
-            const amrex::Array4<amrex::Real const> &state) const;
+    operator()(int ix, int iy, int iz,
+               const amrex::Array4<amrex::Real> &rhs_state,
```
mirenradia, line 88: Can we rename this back to `rhs`? I don't think `rhs_state` makes any sense since it's either the state (i.e. the variables on the grid) or the RHS.

mirenradia: This change should also be made in `CCZ4RHSWithMatter`.

mirenradia: I see now that you've used `rhs` for the `amrex::CellData<...>` you get from this `amrex::Array4<...>`. Maybe it makes sense to make that one `rhs_cell_data` for consistency with `state_cell_data` and then you can make the `Array4<...>` object just `rhs`.

Source/Tagging/ChiExtractionTagger.hpp
```diff
@@ -6,6 +6,7 @@
 #ifndef CHIEXTRACTIONTAGGER_HPP_
 #define CHIEXTRACTIONTAGGER_HPP_
 
+#include "CCZ4Vars.hpp"
```
mirenradia, line 9: I don't think this header is needed here?

KAClough: This file will be deleted when I rebase.

Tests/CCZ4RHSTest/CCZ4RHSTest.hpp
mirenradia: Can you move this header include to `VarsTools-fdf5a7a.hpp`? I believe that's where the compilation error comes from.

Source/Tagging/ChiExtractionTagger.hpp
mirenradia: Ah yes, of course!

Source/CCZ4/CCZ4RHS.hpp
KAClough: I'd prefer to leave it as it is, as having `rhs_cell_data` all through the rhs code is more ugly. I don't see what is wrong with rhs_state being the rhs of the state, and I am sure people will figure it out.

mirenradia: It just seems quite inconsistent and confusing to me.

mirenradia: You're actually already using `rhs_cell_data` in some places!

Source/CCZ4/CCZ4D1Vars.hpp
mirenradia: Sorry, I should have made this clearer. You'll need to define and use another `diff1` function in `FourthOrderDerivatives` that allows you to pass a range of components otherwise we'll get segfaults when it tries to write for vars > `NUM_CCZ4_VARS`.

Source/CCZ4/CCZ4RHS.hpp
KAClough: Yes I know

Source/Grids/DimensionDefinitions.hpp
```diff
@@ -16,11 +16,17 @@ constexpr int GR_SPACEDIM = 3;
 #endif
 
 // NOLINTBEGIN(cppcoreguidelines-macro-usage)
+
+// A function to return the right index for the tensors based on the
+// ordering below 0: T11, 1: T12, 2: T13, 3: T22, 4: T23, 5: T33
+#define VAR_IDX(ivar, i, j) ((ivar) + (i) + (j) + (((i) * (j) != 0) ? 1 : 0))
+
+// A version for where the base reference for the tensor is 0
+#define VAR_IDX0(i, j) VAR_IDX(0, (i), (j))
```
julianakwan, line 25: You could achieve this with using a [single macro with variational arguments](https://stackoverflow.com/questions/11761703/overloading-macro-on-number-of-arguments), but I'm not going to enforce it in this PR 😅 
I'll leave it for another PR and just note it here.

Source/GRTeclynCore/GRAMRLevel.hpp
```diff
@@ -114,9 +114,9 @@ class GRAMRLevel : public amrex::AmrLevel
      * Most examples should not need to override this and instead override
      * pre_tag_cells() and tag_cells()
      */
-    virtual void errorEst(amrex::TagBoxArray &a_tag_box_array, int a_clearval,
-                          int a_tagval, amrex::Real a_time,
-                          int a_n_error_buf = 0, int a_ngrow = 0) override;
+    void errorEst(amrex::TagBoxArray &a_tag_box_array, int a_clearval,
```
julianakwan, line 117: Why was the `virtual` part removed? Was it the linter?

Tests/CCZ4RHSTest/CCZ4RHSTest.cpp
```diff
@@ -106,17 +104,16 @@ void run_ccz4_rhs_test()
...
-            [=] AMREX_GPU_DEVICE(int i, int j, int k)
+            [=] AMREX_GPU_DEVICE(int ix, int iy, int iz)
             {
-                current_ccz4_rhs.compute(i, j, k, current_out_array,
-                                         in_c_array);
-                old_ccz4_rhs.compute(i, j, k, old_out_array, in_c_array);
+                old_ccz4_rhs.compute(ix, iy, iz, old_out_array, in_c_array);
+                current_ccz4_rhs(ix, iy, iz, current_out_array, in_c_array);
 
                 for (int ivar = 0; ivar < NUM_CCZ4_VARS; ++ivar)
```
julianakwan, line 114: Should this be `NUM_VARS`? I'm actually not sure why it isn't `NUM_CCZ4_VARS` throughout...since the RHS in this test shouldn't have any additional variables.

julianakwan (changes requested):
Thanks very much for addressing all the comments in the previous review! I just have 2 minor queries, then I think it is good to merge!

julianakwan:
Just to clarify, I will clean up the wrappers in my PR for the kernel fission/tensor class removal/derivative optimization so I think this looks good to merge once my nitpicky comments have been addressed 😄

Source/Grids/DimensionDefinitions.hpp
KAClough: Ok great, yes, happy to leave the macro magic to you!

Source/GRTeclynCore/GRAMRLevel.hpp
KAClough: Yes right, apparently the virtual is implicit because of something else... I did ask ChatGPT and it agreed with the linter :-)

Tests/CCZ4RHSTest/CCZ4RHSTest.cpp
KAClough: Hmm, I think the StateVariables.hpp that is included from the Common folder does have the scalar variables, so we need to exclude those in the loop where we check the RHS. However, since we only generate initial data for the CCZ4 vars I think we should be fine to change them all to NUM_CCZ4_VARS which I do agree is more consistent. I will amend it to this, and then push and pray.

I also noticed another rogue i,j,k->ix,iy,iz in this file.

julianakwan: I think because we don't look at those variables in this test, it should be okay to specify only `NUM_CCZ4_VARS`.

julianakwan (dismissed):
LGTM

Tests/CCZ4RHSTest/CCZ4RHSTest.hpp
julianakwan: @KAClough I think this is resolved? At least, I don't see the compilation error any more.

Source/CCZ4/CCZ4RHS.hpp
julianakwan: Unfortunately, I agree with @mirenradia on this issue, but we have to move on and merge this PR. So I will resolve this conversation and we can argue about it at a later date.

Tests/CCZ4RHSTest/CCZ4RHSTest.hpp
julianakwan: I have done as @mirenradia requested and moved it into `VarsTools-fdf5a7a.hpp`. Now it is GTM.

KAClough (commented):
This looks good!

SamuelBrady (approved):
LGTM

PR #139, opened 2025-08-19
IO: Add capability to read SmallDataIO files

julianakwan:
Here is my attempt to expand the capability of `SmallDataIO` by adding a reader with more flexibility than reading line by line. Some parts were based on @mirenradia's stand alone `SmallDataIOReader`. 

File format:
- A block of data consists of header lines and then numerical data spaced by at least one blank line (but can be more).  
- Header lines must be prefaced by hashes and there is no limit to the amount of header lines. 
- The last line of the header must be column names for the data that follows. 
- Data is always numerical i.e. no strings allowed in the data columns. 

Reader Features:
- The user can read any combination of data columns by specifying the names of the columns as given in the header e.g. "x" or "time" in a vector of strings or give a range of column numbers (inclusive). 
- Multiple data blocks can be read in at once.  
- Only the MPI rank designated for IO will participate in the actual reading but an utility function is provided to broadcast that data to other ranks. 

I have also added a unit test that writes a file in `SmallDataIO` format and then reads it back in again using the new reader functions.


Source/IO/Make.package
```diff
@@ -1,4 +1,5 @@
-GRTECLYN_CEXE_sources += SmallDataIO.cpp
+GRTECLYN_CEXE_sources += SmallDataIO.cpp
```
mirenradia: I almost missed this whitespace
```suggestion
GRTECLYN_CEXE_sources += SmallDataIO.cpp
```

Source/IO/SmallDataIO.cpp
```diff
@@ -254,10 +271,437 @@ void SmallDataIO::remove_duplicate_time_data(const bool keep_m_time_data)
...
+
+    AMREX_ASSERT_WITH_MESSAGE(
+        m_file, "This function can only be called from the constructor! \n");
+
+    m_file.seekg(0, std::ios::end);
+
+    auto size = m_file.tellg();
+
+    AMREX_ASSERT_WITH_MESSAGE(size < max_file_size,
+                              "File size is too large for SmallDataIO.\n");
```
mirenradia: Do you want to make this an `AMREX_ALWAYS_ASSERT_WITH_MESSAGE` so it's triggered even if not compiling with `DEBUG = TRUE`?

Also, I think it would be nice to prepend error messages with this class name:
```suggestion
    AMREX_ASSERT_WITH_MESSAGE(size < max_file_size,
                              "SmallDataIO: File size is too large.\n");
```

Source/IO/SmallDataIO.cpp
```diff
@@ -254,10 +271,437 @@ void SmallDataIO::remove_duplicate_time_data(const bool keep_m_time_data)
...
+
+    std::string file_contents(size,
+                              '\0'); // construct string to stream size
+    m_file.seekg(0);
+    m_file.read(file_contents.data(), size);
+    AMREX_ASSERT_WITH_MESSAGE(!file_contents.empty(), "File contents of " +
+                                                          m_filename +
+                                                          " could not be read");
```
mirenradia: Same comment as above re `ALWAYS_ASSERT`.

Let's also prepend the error message with `SmallDataIO` (formatting almost
certainly wrong in this suggestion):
```suggestion
    AMREX_ASSERT_WITH_MESSAGE(!file_contents.empty(), "SmallDataIO: "
                              "File contents of " + m_filename +
                               " could not be read");
```

Source/IO/SmallDataIO.cpp
```diff
@@ -254,10 +271,437 @@ void SmallDataIO::remove_duplicate_time_data(const bool keep_m_time_data)
...
+        }
+
+        std::istringstream file_stream(m_file_contents);
+        skip_ahead(file_stream, m_file_structure.num_header_rows[a_block],
+                   a_block);
+
+        for (int irow = 0; irow < m_file_structure.num_data_rows[a_block];
+             ++irow)
+        {
+            double discard = 0.0;
```
mirenradia: Can you move this variable instantiation into the `else` block below so that its scope is minimised.

I think the performance difference will be minimal as the compiler will probably just optimize away the variable anyway.

Source/IO/SmallDataIO.cpp
```diff
@@ -254,10 +271,437 @@ void SmallDataIO::remove_duplicate_time_data(const bool keep_m_time_data)
...
+
+        std::istringstream file_stream(m_file_contents);
+
+        skip_ahead(file_stream, m_file_structure.num_header_rows[a_block],
+                   a_block);
+
+        for (int irow = 0; irow < m_file_structure.num_data_rows[a_block];
+             ++irow)
+        {
+            double discard = 0.0;
```
mirenradia: Same comment as for the `discard` variable in the above function.

Source/IO/SmallDataIO.cpp
```diff
@@ -254,10 +271,437 @@ void SmallDataIO::remove_duplicate_time_data(const bool keep_m_time_data)
...
+
+            // This is a bit slow but returns the columns in the order that the
+            // user requested them
+
+            for (int icolumn = 0;
+                 icolumn < m_file_structure.num_data_columns[a_block];
+                 ++icolumn)
+            {
+                auto const column_ptr = std::find(
+                    column_numbers.begin(), column_numbers.end(), icolumn);
```
mirenradia: As discussed, let's try to avoid doing this `std::find` for every row.

We can sort the `column_numbers` vector and also store a `std::map` from the column number to the request number to ensure it's returned in the same order to the user.

Source/IO/SmallDataIO.cpp
```diff
@@ -355,4 +801,32 @@ double SmallDataIO::get_default_coords_epsilon()
...
+
+    for (auto &column : data)
+    {
+        column.resize(nrows);
+        amrex::ParallelDescriptor::Bcast(
+            column.data(), nrows,
+            amrex::ParallelDescriptor::IOProcessorNumber());
+    }
+
+    amrex::ParallelDescriptor::Barrier();
```
mirenradia: Is this barrier necessary?

Source/IO/SmallDataIO.hpp
```diff
@@ -135,8 +172,55 @@ class SmallDataIO
     void get_specific_data_line(std::vector<double> &a_out_data,
                                 const double a_coord);
 
+    // Set structure if known already (e.g. same as another file already
+    // determined)
+    void set_file_structure(const file_structure_t &a_file_structure);
```
mirenradia, line 195: Currently this function is a bit useless as `determine_file_structure` is called
in the constructor.

I wonder if it's worth adding a separate mode e.g. `READ_NO_DETERMINE_STRUCTURE`
(or something less verbose) where the file structure is not determined in the
constructor but can be set manually?

Tests/SmallDataIOTest/Make.package
mirenradia, line 1: Why do you have `SmallDataIO.hpp` and `SmallDataIO.cpp` in this file? I think they should be picked up anyway from `Source/IO/Make.package`?

Tests/SmallDataIOTest/SmallDataIOTest.cpp
```diff
@@ -0,0 +1,181 @@
...
+    for (int i = 0; i < vector_1.size(); ++i)
+    {
+        if (std::abs(vector_1[i] - vector_2[i]) > err_tol)
+        {
+            return false;
+        }
+    }
+
+    return true;
+}
```
mirenradia: What do you think about moving the doctest `CHECK` inside this function and doing something like
```cpp
INFO("i = " << i)
CHECK(vector_1[i] == doctest::Approx(vector_2[i]).epsilon(err_tol))
```
inside the loop?

If it fails, it will provide more information about which line is causing the failure.

Tests/SmallDataIOTest/SmallDataIOTest.cpp
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        amrex::Gpu::streamSynchronize();
+
+        static constexpr int data_precision = 10;
+        static const amrex::Real err_tol = std::pow(10., -1.0 * data_precision);
+
+        // Test the file writer: write out the random co-oridinates
+
+        test_small_data_io_writer(write_data, data_precision);
```
mirenradia: Just a minor typo/formatting.
```suggestion
        // Test the file writer: write out the random coordinates
        test_small_data_io_writer(write_data, data_precision);
```

mirenradia (changes requested):
Thanks for this PR, @julianakwan. It looks great! :smile: I particularly like the test and that you replaced all of the old `m_rank == 0`s with `amrex::ParallelDescriptor::IOProcessor()`.

I have a few minor comments but I think the biggest one is moving the `std::find` out of the rows loop as we discussed in person.

Source/IO/SmallDataIO.hpp
julianakwan: I have changed this so that now you can call `SmallDataIO` in two different ways: 
```
SmallDataIO test_reader("test_000000");
```
as per usual when you don't know the file structure or
```
const SmallDataIO::file_structure_t test_file_structure{1, {0}, {1}, {100}, {4}};
SmallDataIO test_reader("test_000000", &test_file_structure);
```
This is from the unit test so there is an example for users.

Source/IO/SmallDataIO.cpp
```diff
@@ -254,10 +301,443 @@ void SmallDataIO::remove_duplicate_time_data(const bool keep_m_time_data)
...
+                std::string error_message{
+                    "SmallDataIO: " + column_names[icolumn] +
+                    " could not be read. Please double check "
+                    "your inputs to SmallDataIO::get_columns"};
+                amrex::Abort(error_message);
+            }
+        }
+
+        const int ncols = m_file_structure.num_data_columns[a_block];
+        std::vector<bool> truth_table(ncols, false);
```
mirenradia: Can we give this a more descriptive name than `truth_table`? Maybe something like `file_columns_requested`?

It would also be worth write a brief comment explaining that we create this object to save having to call `std::map::find` during the loop below.

Source/IO/SmallDataIO.cpp
```diff
@@ -254,10 +301,443 @@ void SmallDataIO::remove_duplicate_time_data(const bool keep_m_time_data)
...
+        std::istringstream file_stream(m_file_contents);
+
+        skip_ahead(file_stream, m_file_structure.num_header_rows[a_block],
+                   a_block);
+
+        double discard = 0.0;
+        for (int irow = 0; irow < m_file_structure.num_data_rows[a_block];
+             ++irow)
+        {
+            for (int icolumn = 0;
```
mirenradia: It would be good if the loop variable name followed the conventions used above (i.e. `iheader` for the columns in the file and `icolumn` for the user-requested columns). I guess `iheader` feels a bit wrong as we're not iterating through the header but I think it should be consistent even if that means renaming `iheader` above.

mirenradia (changes requested):
Thanks Juliana for making the changes from the previous review.

I think we're almost there. Just two very small comments about variable names.

Source/IO/SmallDataIO.hpp
mirenradia: This looks good!

Source/IO/SmallDataIO.cpp
julianakwan: I changed it to `columns_requested` and added the comment: 

```
        // To avoid calling std::map::find for each row of data,                                            
        // we define a bool for each column (in the original file ordering) so                              
        // we know if that value needs to be saved or not.  
```

Source/IO/SmallDataIO.cpp
julianakwan: I think this is _more_ confusing 😅 Because the loop on L584 is over all columns in the header so changing it to `irow` doesn't really make sense. The loop over `irow` above is over the data rows which is a different dimension. I could change it to `idata_row`?

mirenradia: I'm now confused by your comment. I didn't mention `irow` and was referring to the use of `icolumn`.

julianakwan: Ah ok I see what you mean now. I changed `iheader` to `file_column` and `icolumn` to `input_column`.

mirenradia (approved):
LGTM

PR #140, opened 2025-08-20
Move where NaNs are checked for

### LINKED ISSUE #100: Move where NaNs are checked for
In the BinaryBH example, NaNs are checked for in `BinaryBHLevel::specificAdvance()`. We should move this to `GRAMRLevel::advance()` or `GRAMRLevel::post_timestep()` as it's not problem-specific.

Moved from #92.

gokmenkilic:
This PR includes moving the nanchecks from `BinaryBHLevel::specificAdvance()` to `GRAMRLevel::post_timestep()` #100

### DIFF
--- a/Source/GRTeclynCore/GRAMRLevel.cpp
+++ b/Source/GRTeclynCore/GRAMRLevel.cpp
@@ -199,6 +199,15 @@ void GRAMRLevel::post_timestep(int /*iteration*/)
 
         FourthOrderInterpFromFineToCoarse(S_crse, 0, NUM_VARS, S_fine, ratio);
     }
+    if (simParams().nan_check)
+    {
+        amrex::MultiFab &state_new = get_new_data(State_Type);
+        if (state_new.contains_nan(0, state_new.nComp(), amrex::IntVect(0),
+                                   true))
+        {
+            amrex::Abort("NaN in GRAMRLevel::post_timestep");
+        }
+    }
 
     specificPostTimeStep();
 }
--- a/Examples/BinaryBH/BinaryBHLevel.cpp
+++ b/Examples/BinaryBH/BinaryBHLevel.cpp
@@ -56,19 +56,7 @@ void BinaryBHLevel::specificAdvance()
                            PositiveChiAndLapse()(cell);
                        });
 
-    // Check for nan's
-    if (simParams().nan_check)
-    {
-        if (S_new.contains_nan(0, S_new.nComp(), amrex::IntVect(0), true))
-        {
-            amrex::Abort("NaN in specificAdvance");
-        }
-    }
-    else
-    {
-        // stream sync already present in nan_check so only need this here
-        amrex::Gpu::streamSynchronize();
-    }
+    
 }
 
 // This initial data uses an approximation for the metric which
--- a/Examples/BinaryBH/BinaryBHLevel.cpp
+++ b/Examples/BinaryBH/BinaryBHLevel.cpp
@@ -55,8 +55,6 @@ void BinaryBHLevel::specificAdvance()
                            TraceARemoval()(cell);
                            PositiveChiAndLapse()(cell);
                        });
-
-    
 }
 
 // This initial data uses an approximation for the metric which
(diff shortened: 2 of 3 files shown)


Source/GRTeclynCore/GRAMRLevel.cpp
```diff
@@ -199,6 +199,15 @@ void GRAMRLevel::post_timestep(int /*iteration*/)
 
         FourthOrderInterpFromFineToCoarse(S_crse, 0, NUM_VARS, S_fine, ratio);
     }
+    if (simParams().nan_check)
+    {
+        amrex::MultiFab &S_new = get_new_data(State_Type);
+        if (S_new.contains_nan(0, S_new.nComp(), amrex::IntVect(0), true))
```
mirenradia: I think it would be good to add a space before this block and rename `S_new` -> `state_new` (as we will do elsewhere in other PRs).

The following suggestion is almost certainly not clang-formatted correctly but you get the idea:

```suggestion

    if (simParams().nan_check)
    {
        amrex::MultiFab &state_new = get_new_data(State_Type);
        if (state_new.contains_nan(0, state_new.nComp(), amrex::IntVect(0), true))
```

Source/GRTeclynCore/GRAMRLevel.cpp
```diff
@@ -199,6 +199,15 @@ void GRAMRLevel::post_timestep(int /*iteration*/)
...
+        {
+            // Abort with a message indicating the new, general location
```
mirenradia: I don't think this comment is necessary.
```suggestion
```

mirenradia (changes requested):
Thanks for the PR, Gokmen. In addition to the suggestions below and fixing the clang-formatting, please could you rewrite the git history using an interactive rebase to do the following:
1. Fix the typos in your commit messages.
2. Squash the first two commits together.
3. Add a more descriptive commit message for the final commit e.g.
   ```
   BinaryBH: Enable NaN check in params_test
   ```
   In general, I like to try and prepend the commit message subject by the area of the code or class that is being modified (as for the example above) but sometimes it's difficult to do.

Source/GRTeclynCore/GRAMRLevel.cpp
```diff
@@ -199,6 +199,19 @@ void GRAMRLevel::post_timestep(int /*iteration*/)
...
+    {
+        amrex::MultiFab &state_new = get_new_data(State_Type);
+        if
+
+            (state_new.contains_nan(0, state_new.nComp(), amrex::IntVect(0),
+                                    true))
+
+        {
+
+            amrex::Abort("NaN in GRAMRLevel::post_timestep");
```
mirenradia: I'm a bit surprised clang-format was OK with all of this unnecessary whitespace but let's get rid of it:
```suggestion
        if (state_new.contains_nan(0, state_new.nComp(), amrex::IntVect(0),
                                   true))
        {
            amrex::Abort("NaN in GRAMRLevel::post_timestep");
```

mirenradia (changes requested):
I'm not sure what happened since your previous iteration but there now seems to be a lot of excess whitespace. Would you be able able to fix it (and squash the change into your first commit)?

Source/GRTeclynCore/GRAMRLevel.cpp
gokmenkilic: I think, I messed up something while rebasing. The PR should be clean now, just faling because of Tursa is down.

mirenradia: I think the whitespace is still there. We don't use Tursa for CI but CSD3. I believe it is failing because some of the login nodes are down for maintenance.

mirenradia (changes requested):
Unfortunately I don't think the whitespace issues were resolved from my previous review.

I'm not sure what's happened but in your second commit ([Enable NaN check in params_test](https://github.com/GRTLCollaboration/GRTeclyn/pull/140/commits/ea199472f055ebff4495b10f63282e2871795c68)) there are now whitespace changes in `BinaryBHLevel.cpp` too.

mirenradia (approved):
LGTM (GitLab CI failing as CSD3 is down today)

Source/GRTeclynCore/GRAMRLevel.cpp
mirenradia: It's gone now!

PR #148, opened 2025-10-03
CI: Limit running by which paths are changed

mirenradia:
This PR limits when some of the CI workflows run. In particular, I have added conditions for the workflows that build code as these take the most time.

In the case of GitHub actions, I have used the `paths-ignore` option to explicitly ignore changes in certain paths (i.e. a blocklist). I think this is the best solution as if we add new paths that we would like to trigger CI for, we don't have to change anything. The disadvantage is that we might need to add more paths to ignore in the future.

In the case of GitLab CI, I have used `rules:changes` to explicitly say which paths will trigger the CI if there are changes (i.e. an allowlist) This is because blocklists as described above don't really work if commits contain both paths we want to allow and paths we want to ignore (see [here](https://gitlab.com/gitlab-org/gitlab/-/issues/198688) for details).

I have also fixed the formatting of the CI YAML files so that the pre-commit checks introduced in #124 pass on these files.

Finally, I have removed the deletion of the `QOS_INTR_LOCK_FILE` as I think it was causing concurrent GitLab pipelines to sometimes fail and I don't think we need to delete it either. If the `srun` command fails, I think the file lock should be released.

### DIFF
--- a/.github/workflows/check-changes.yml
+++ b/.github/workflows/check-changes.yml
@@ -1,0 +1,31 @@
+name: Check for changes to Source files
+
+# This allows us to exclude certain GitHub actions from running
+# if, say, only changes to the documentation are made.
+# Simply add this as the first step in your jobs and check the
+# return value in subsequent steps. It will return true if any
+# of the files in the 'has_changes' variable is true
+
+on:
+  workflow_call:
+    outputs:
+      has_changes:
+        value: ${{ jobs.check-changes.outputs.has_changes }}
+
+jobs:
+  check-changes:
+    runs-on: ubuntu-24.04
+    outputs:
+      has_changes: ${{ steps.filter.outputs.has_changes }}
+    steps:
+      - uses: actions/checkout@v5
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            has_changes:
+              - 'Source/**'
+              - '.github/**'
+              - 'GNUmakefile'
+              - 'Examples/**'
+              - 'Tools/GNUMake'
--- a/.github/workflows/check-changes.yml
+++ b/.github/workflows/check-changes.yml
@@ -9,23 +9,36 @@ name: Check for changes to Source files
 on:
   workflow_call:
     outputs:
-      has_changes:
-        value: ${{ jobs.check-changes.outputs.has_changes }}
+      has_core_changes:
+        value: ${{ jobs.check-changes.outputs.has_core_changes }}
+      has_example_changes:
+        value: ${{ jobs.check-changes.outputs.has_example_changes }}
+      has_unit_test_changes:
+        value: ${{ jobs.check-changes.outputs.has_unit_test_changes }}
 
 jobs:
   check-changes:
     runs-on: ubuntu-24.04
     outputs:
-      has_changes: ${{ steps.filter.outputs.has_changes }}
+      has_core_changes: ${{ steps.filter.outputs.has_core_changes }}
+      has_example_changes: ${{ steps.filter.outputs.has_example_changes }}
+      has_unit_test_changes: ${{ steps.filter.outputs.has_unit_test_changes }}
     steps:
       - uses: actions/checkout@v5
       - uses: dorny/paths-filter@v3
         id: filter
         with:
+          base: ${{ github.event_name == 'pull_request' && '' || github.ref }}
           filters: |
-            has_changes:
+            has_core_changes: &core
               - 'Source/**'
               - '.github/**'
               - 'GNUmakefile'
+              - 'Tools/GNUMake/**'
+            has_example_changes:
+              - *core
               - 'Examples/**'
-              - 'Tools/GNUMake'
+            has_unit_test_changes:
+              - *core
+              - 'Test/**'
+              - 'external/doctest/**'
--- a/.github/workflows/gpu-build.yml
+++ b/.github/workflows/gpu-build.yml
@@ -4,10 +4,6 @@ name: GPU Build
 
 on: [push, pull_request]
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
-
 jobs:
   check-changes:
     uses: ./.github/workflows/check-changes.yml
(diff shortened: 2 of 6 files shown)


mirenradia:
You can see an example of fewer CI workflow running on f422eb1fa7cc024c2ed2055bb425c60a45ce56d4.

mirenradia:
You may notice I have added a `pull_request` trigger to all of the GitHub actions workflows I've modified (previously it was just `push`). This is to allow for PRs where the final pushed commit has only ignored changes but there are relevant changes in earlier commits that we'd like to test. Unfortunately, this does mean that these workflows will run twice for relevant changes in PRs. I'm not sure if there's a way to avoid this.

julianakwan (changes requested):
Thanks for this much needed PR! I think we discussed this before, but perhaps some of the duplicated content in `gpu-build.yml`, `regression-tests.yml` and `unit-tests.yml` could be moved into a common file? (I know they are not completely the same, only mostly the same...). We might want to add new directories to the list of excluded paths one day and currently we would have to make the changes to each of these files separately.
Also I think it's time we had a common file - it would make refactoring the CI easier in the future as there are some other duplicated steps that could go in there as well (e.g. setting up the cache and dependencies, but this is for another PR).

mirenradia:
> Thanks for this much needed PR! I think we discussed this before, but perhaps some of the duplicated content in `gpu-build.yml`, `regression-tests.yml` and `unit-tests.yml` could be moved into a common file? (I know they are not completely the same, only mostly the same...). We might want to add new directories to the list of excluded paths one day and currently we would have to make the changes to each of these files separately. Also I think it's time we had a common file - it would make refactoring the CI easier in the future as there are some other duplicated steps that could go in there as well (e.g. setting up the cache and dependencies, but this is for another PR).

Unfortunately, after looking at several options, I think none of them will work for the `paths-ignore` key so I think we will have to accept the duplication.

mirenradia:
I've rebased this onto `develop` following the merge of #121. It's ready to be re-reviewed.

julianakwan:
Hi @mirenradia - I looked into this some more and have made some suggested changes in the branch `enhancement/limit_ci_jk_suggest`. 

I use the GitHub Action [here](https://github.com/dorny/paths-filter) to look for changes to files in certain directories for which we want the CI to run (e.g. `Source/*`). I moved this into a separate job in the file `check-changes.yml` and then call this as the first step for the regression test, unit test etc. The subsequent step then uses the `needs:` and `if:` keywords to check the output variables from `check-changes.yml` before the running the rest of the step. 

I hope this is acceptable :) 

NB: I learnt about the `paths-filter` action from the AMReX GitHub - it is what they use to stop the CI from running when they push changes to their documentation.

julianakwan (changes requested):
Please see the changes I suggested in the branch `enhancement/limit_ci_jk_suggest`. Thanks!

mirenradia:
I'm happy with your changes but I'm wondering if we can change the trigger to just `push` (i.e. disable the `pull_request` trigger) and then set the `base` to `develop` so changes are compared with `develop` and we don't need to run workflows twice in PRs.

julianakwan:
> I'm happy with your changes but I'm wondering if we can change the trigger to just `push` (i.e. disable the `pull_request` trigger) and then set the `base` to `develop` so changes are compared with `develop` and we don't need to run workflows twice in PRs.

I think that will work as long as you're happy for regression test etc to always run, even if you've only changed the documentation for that particular commit. That was my reasoning behind setting the value of `base:` to the previous commit as some commits within a branch will only change the documentation for the parts of the code that have been updated (as people should!).

mirenradia:
Actually I just tried pushing two commits to `test/limit_ci_jk_suggest`, the first (earliest) one had a change on a relevant source file and the second (latest) one just had a change in the README. With your current configuration, `check-changes` successfully recognised that there were changes in the push on relevant files (even though there weren't any in the last commit) so the CI was triggered.

I think this means we can leave your workflow as it is and just remove the `pull_request` trigger.

mirenradia:
> Actually I just tried pushing two commits to `test/limit_ci_jk_suggest`, the first (earliest) one had a change on a relevant source file and the second (latest) one just had a change in the README. With your current configuration, `check-changes` successfully recognised that there were changes in the push on relevant files (even though there weren't any in the last commit) so the CI was triggered.
> 
> I think this means we can leave your workflow as it is and just remove the `pull_request` trigger.

Actually I'm less sure about this now. Our concurrency configuration means that older workflows will be cancelled if they're still running and there's a new push.

mirenradia:
OK, I think I've solved the concurrency problem by moving it from the workflow to the job level and renaming the groups appropriately.

I've tested this on the `test/limit_ci` branch. It works as I would expect but I've now realised that maybe we don't want to get rid of the `pull_request` trigger since if the last commit on a PR branch is e.g. just modifying the README, the CI can pass on that commit and I think it will say "all checks have passed" on the PR page but they can still fail on earlier commits. I guess we don't want to rely on reviewers checking that the regression tests/GPU build/unit tests workflows have passed on earlier commits.

mirenradia:
I've re-added the `pull_request` trigger to the GitHub workflows. For the GitLab pipeline, I've also moved the `QOS_INTR_LOCK_FILE` on CSD3 to RDS storage so that hopefully all login nodes can see this locked file coherently.

I should point out that the check-changes workflow uses a positive include filter so if we add new paths that we want changes on to trigger running these GitHub workflows, we will need to remember to modify this workflow accordingly. I probably would have preferred an ignorelist for the reasons I gave in my original PR description but I don't think there's a way of doing this using the paths-filter action.

julianakwan:
> I should point out that the check-changes workflow uses a positive include filter so if we add new paths that we want changes on to trigger running these GitHub workflows, we will need to remember to modify this workflow accordingly. I probably would have preferred an ignorelist for the reasons I gave in my original PR description but I don't think there's a way of doing this using the paths-filter action.

I chose to use a positive include filter because I thought it resulted in fewer paths being checked so the step would be tidier. 

If you like, you can use the exclusion syntax, but you have to use the `every` keyword as well so that all the conditions are satisfied. For example: 

```
- uses: dorny/paths-filter@v3
      id: filter
      with:
        # This makes it so that all the patterns have to match a file for it to be
        # considered changed. Because we have the exclusions for .jpeg and .md files
        # the end result is that if those files are changed they will be ignored
        # because they don't match the respective rules excluding them.
        #
        # This can be leveraged to ensure that you only build & test software changes
        # that have real impact on the behavior of the code, e.g. you can set up your
        # build to run when Typescript/Rust/etc. files are changed but markdown
        # changes in the diff will be ignored and you consume less resources to build.
        predicate-quantifier: 'every'
        filters: |
          backend:
            - 'pkg/a/b/c/**'
            - '!**/*.jpeg'
            - '!**/*.md'
```

mirenradia:
OK, let's leave things as they are. I guess at least it's a consistent approach between the GitLab and GitHub CI configurations.

julianakwan (approved):
Thanks for the changes! I like the logic in the `base` keyword to cover both cases of changes relative to the last commit and also if the action is run as a result of a PR. 

Looks good to merge to me :)

PR #158, opened 2026-01-29
fix: update Array4 stride access for compatibility with latest AMReX API

Nikchik-coder:
Summary:

This PR fixes a compilation error caused by recent breaking changes in the AMReX Array4 (and ArrayND) API. The direct members nstride, jstride, and kstride have been deprecated/removed in the latest AMReX versions.

Changes:
Replaced deprecated direct member access (state.jstride, etc.) with the official template-based accessor: state.template get_stride<N>().
Updated memory offset calculations to use the new API across the core derivative engine and example implementations.
Applied project-wide code formatting using clang-format to ensure consistency with the repository's style guide.

Files modified:
Source/Grids/FourthOrderDerivatives.hpp: Main finite difference engine.
Examples/KleinGordon/KleinGordonRHS.impl.hpp: Scalar field example implementation.
Tests/CCZ4RHSTest/FourthOrderDerivatives-fdf5a7a.hpp: Regression test support file.
Testing:
Verified successful local build of the BinaryBH example using the latest AMReX.
All 60 CI checks (Unit Tests, Regression Tests, GPU Builds, and Linting) passed successfully on GitHub Actions.

### DIFF
--- a/Examples/KleinGordon/KleinGordonRHS.impl.hpp
+++ b/Examples/KleinGordon/KleinGordonRHS.impl.hpp
@@ -22,13 +22,14 @@ KleinGordonRHS<model_t, deriv_t>::compute(
     const auto *input_ptr_ijk = input.ptr(i, j, k);
     amrex::Array1D<amrex::Real, 0, AMREX_SPACEDIM>
         d2phi{}; // no cross second order derivatives needed
-    amrex::Array1D<int, 0, AMREX_SPACEDIM> strides{AMREX_D_DECL(
-        1, static_cast<int>(input.jstride), static_cast<int>(input.kstride))};
+    amrex::Array1D<int, 0, AMREX_SPACEDIM> strides{
+        AMREX_D_DECL(1, static_cast<int>(input.stride.a[0]),
+                     static_cast<int>(input.stride.a[1]))};
 
     FOR (i)
     {
-        d2phi(i) =
-            m_deriv.diff2(input_ptr_ijk + c_phi * input.nstride, 0, strides(i));
+        d2phi(i) = m_deriv.diff2(input_ptr_ijk + c_phi * input.stride.a[2], 0,
+                                 strides(i));
     }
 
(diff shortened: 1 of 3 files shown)


mirenradia (changes requested):
Hi @Nikchik-coder, thanks for the PR that fixes this issue.

Please could you make the following changes:
* Remove all formatting of unrelated files. We try to keep each PR to making a single self-contained change rather than touching lots of unrelated parts of the codebase. Furthermore, I think you may have been using the wrong version of clang-format as some of the files you have modified should already be correctly formatted. You can [set up pre-commit](https://github.com/GRTLCollaboration/GRTeclyn/wiki/Using-Clang-Format#setting-up-the-pre-commit-hook) to make sure it's correct.
* Remove use of the `.template ` disambiguator keyword. Since we're using C++17, this hopefully shouldn't be necessary (though some compilers may be fussier than others) and I think it reduces readability.

Since this fix should only affect a few files, please could you do an interactive rebase to edit the history such that the changes are made in a single commit.

I'll approve the CI workflows after you've made these changes.

P.S. May I ask why you have opened a PR to this repo? From your GitHub profile, it doesn't appear as if you work in numerical relativity, hence the question.

Nikchik-coder:
Hi, I will certainly include those fixes.

Regarding your question about the motivation for this PR: I previously used GRChombo for a personal project and, having an MS in Physics, I am quite passionate about this field.

I recently decided to try GRTeclyn because I am a big fan of GPU-accelerated computing. However, I encountered a build failure with my compiler, which led to this PR. I believe these changes will be helpful for other users as well.

Thank you for all your work on this project; it is truly amazing!

mirenradia (approved):
Thanks for making the previous changes I requested.

I think the lint ignore CI workflow is failing because of the permissions of the GitHub token (almost all of our PRs are from branches in this repo rather than forks) so I think it can be safely ignored.

PR #166, opened 2026-03-04
Style: Fix copyright headers and restore CI job to check it

mirenradia:
This PR fixes all the copyright headers in C++ source files and also modifies the `check_copyright_headers` and `add_copyright_headers` script to ensure there is a blank line between the copyright header and the next non-empty line.

It turns out I accidentally removed the CI job that ran `check_copyright_headers` in #57 so I have now restored it.


Tools/CopyrightHeaders/add_copyright_headers
```diff
@@ -17,9 +17,19 @@ copyright_header=$(cat $1)
 num_lines=$(echo "$copyright_header" | wc -l)
 
 for f in $(find $REPO_ROOT -type f -name "*.[h,c]pp" -not -path "$REPO_ROOT/External/*"); do
```
julianakwan, line 19: Should the copyright header be applied to other types of files as well? Sometimes we have shell scripts and I can imagine having example Python scripts one day for plotting and/or analysis etc.

julianakwan (changes requested):
Thanks a lot for restoring the copyright header checker! I think I was responsible for most of the missing headers :)

I just had the one query that I thought might be worth having a conversation about but I am happy to shelve this for another day if you don't want to address it right now.

Tools/CopyrightHeaders/add_copyright_headers
mirenradia: I don't feel strongly whether we add to other types of files or not at some point but this will complicate things a bit as the comment syntax will vary between types. For this PR, I think we should just leave it to our C++ source files as we have always done.

julianakwan (approved):

PR #167, opened 2026-03-04
Replace GRChombo with GRTeclyn in end printout

mirenradia:
Even though the ScalarField example has yet to be ported, I have made the change there as well for consistency.

This has annoyed me for quite a while...

### DIFF
--- a/Examples/BinaryBH/Main_BinaryBH.cpp
+++ b/Examples/BinaryBH/Main_BinaryBH.cpp
@@ -82,12 +82,12 @@ int main(int argc, char *argv[])
 
     if (status == 0)
     {
-        amrex::Print() << "GRChombo finished."
+        amrex::Print() << "GRTeclyn finished."
                        << "\n";
     }
     else
     {
-        amrex::Print() << "GRChombo failed with return code " << status << "\n";
+        amrex::Print() << "GRTeclyn failed with return code " << status << "\n";
     }
 
     mainFinalize();
--- a/Examples/ScalarField/Main_ScalarField.cpp
+++ b/Examples/ScalarField/Main_ScalarField.cpp
@@ -55,9 +55,9 @@ int main(int argc, char *argv[])
     int status = runGRTeclyn(argc, argv);
 
     if (status == 0)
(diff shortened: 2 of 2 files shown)


julianakwan (commented):
Thanks @mirenradia ! 

It would be really helpful if we could also print out the Git SHA like in AMReX, especially when benchmarking. 

Do you think you could implement something like this: 
1) Add a print statement to main_XXX.cpp [like this](https://github.com/AMReX-Codes/amrex/blob/6c3aa138b2b1635c99b72eddedb0f058ae26d3ef/Src/Base/AMReX_Version.cpp#L11)

2) Populate `VERSION_NUMBER` in Make.rules with the SHA [like this](https://github.com/AMReX-Codes/amrex/blob/6c3aa138b2b1635c99b72eddedb0f058ae26d3ef/Tools/GNUMake/Make.defs#L411)

🙏 I've been meaning to do it myself for a while...

mirenradia:
> Thanks @mirenradia !
> 
> It would be really helpful if we could also print out the Git SHA like in AMReX, especially when benchmarking.
> 
> Do you think you could implement something like this:
> 
>     1. Add a print statement to main_XXX.cpp [like this](https://github.com/AMReX-Codes/amrex/blob/6c3aa138b2b1635c99b72eddedb0f058ae26d3ef/Src/Base/AMReX_Version.cpp#L11)
> 
>     2. Populate `VERSION_NUMBER` in Make.rules with the SHA [like this](https://github.com/AMReX-Codes/amrex/blob/6c3aa138b2b1635c99b72eddedb0f058ae26d3ef/Tools/GNUMake/Make.defs#L411)
> 
> 
> 🙏 I've been meaning to do it myself for a while...

I agree this is a good idea and something I would like as well. However, I don't think it belongs in this PR so I'll open an issue. It would probably make sense if this went in `mainSetup()` too so that it's not necessary to implement in every example's `main()` or `runGRTeclyn()`.

julianakwan (approved):
Ok, I'm happy for the printing of the Git SHA to go in a different PR.

PR #173, opened 2026-04-08
Enhancement/re organising random fields class

the-florist:
This branch contains a partial version of my GRTeclyn example, which initialises and evolves stochastic scalar and gravitational-wave fields during inflation. The main alterations to GRTeclyn made in this example are found in the Examples/ScalarFields folder, which contain five new hpp or hpp/impl.hpp file pairs that together perform this stochastic initialisation, extract cosmologically-relevant observables and perform tests on the gravitational-wave field as the simulation evolves:
- RandomFieldInit: creates and stores the initial conditions inside GRTeclyn's state object,
- InflationExtraction: extracts the scalar and tensor perturbations, finds their power spectra, etc. and also contains a derive() function to store derived variables associated with this example in plotfiles,
- InflaitonConfig: creates and tracks a struct tracking parameters and flags used in both the init and extraction routines, 
- InflationUtils: stores const objects that are used in both init and extraction,
- TensorTests: performs test and consistency checks on the gravitational waves during initialisation and extraction.

In order to run this example, one can set the parameter read_from_STOIIC=0, which will initialise the scalar and tensor fields according to an approximate power spectrum. One can also run STOIIC_GR, Yoann Launay's code, to create a params.txt file containing a more precise version of the power spectrum for scalar and tensor fields. These spectra are then read in inside the RandomFieldInit class when read_from_STOIIC=1.


the-florist:
Apologies, meant to make this internal to my fork...

Issue #149, opened 2025-10-03
Switch to AMReX geometry parameters and use negative extents


mirenradia:
This is heavily related to #1 but I thought I'd move discussions on it into a separate issue.

The relevant AMReX geometry parameters are:

* `geometry.prob_lo`
* `geometry.prob_hi`
* `amr.n_cell`

In addition to simplifying the horrible mess of `Ni`, `N`, `N_full`, etc. parameters, we also want to shift the position of the grid such that the centre is at the origin (i.e. $(0,\\,0,\\,0)$ ). Assuming there is no reflective symmetry, if want the domain to have side lengths $L_i$ for $i=1,2,3$, the low end will be $(-L_1/2,\\,-L_2/2,\\,-L_3/2)$ and the high end will be $(L_1/2,\\,L_2/2,\\,L_3/2)$.

I have made a start on this on the [`enhancement/negative_extents`](https://github.com/GRTLCollaboration/GRTeclyn/tree/enhancement/negative_extents) branch but it probably makes sense for these changes to be made either as part of work on #1 or after that is done.

tamaraevst:
A reminder for future selves to change the way we determine whether we are at the low/high boundary in `Lagrange.hpp` (see [here](https://github.com/GRTLCollaboration/GRTeclyn/pull/174#discussion_r3550196661) and [here](https://github.com/GRTLCollaboration/GRTeclyn/blob/b0c8781f413a491080fed4741451cf849c7470ee/Source/ParticleInterpolator/Lagrange.hpp#L55)). Current logic works because the low boundary is at (0,0,0); this issue will likely change this.

Issue #107, opened 2025-03-19
BinaryBH: Clean up parameter files


mirenradia:
We currently have 6 parameter files when probably only need 3 or 4 max. Let's get rid of some of them.

PR #174, opened 2026-04-13
Port extraction class to ParticleInterpolator

### LINKED ISSUE #16: Port Surface Extraction capabilities

tamaraevst:
This is the first draft for porting all of the `Extraction*` classes to GRTeclyn and addressing issue #16. Key features:

- The class structure of all extraction files being ported here inherit everything from GRChombo. Most importantly, the inheritance of `SphericalExtraction` from `SurfaceExtraction` is the same.
- All of the extraction is fully integrated with the `ParticleInterpolator`.
- I have tested the GW extraction on a $$q=1$$ non-spinning BBH head-on collision using initial separation of $$d = 10M$$ and no boost. Comparisons across different resolutions, extraction radii within GRTeclyn and medium and high resolutions between GRChombo and GRTeclyn are attached below. In summary, within GRTeclyn I get convergence of around 2nd order on the radiated energy. When comparing to GRChombo, I get closer results with GRTeclyn at medium and high resolutions. My resolutions correspond to 1/64, 1/96 and 1/128 on the finest levels for low, med and high resolutions, respectively. I forgot to put y-labels, but I am showing the (22)-mode only. The agreement/comparison for the (20)-mode looks qualitatively similar. A sample parameter file (at medium resolution) I used is also attached in file [params.txt](https://github.com/user-attachments/files/26689224/params.txt).
- The test for `SphericalExtraction` has also been ported. We treat spherical harmonics as state variables there. This choice was deliberate as the above was an "indirect" tests for derived-like variables.

extraction_radii
grteclyn_resolutions
grteclyn_grchombo_comparison

I have a couple of comments in regards to the current structure, which I find sub-optimal:

- `SphericalExtraction` parameters are stored in `SimulationParametersBase`. Because our `ParticleInterpolator` is templated over the number of components we are interpolating over, `SphericalExtraction` needs to inherit the exact same template. Now, this means in `SimulationParametersBase` we need to instate this number of components. Currently, it is hard-coded as a number. One would have hoped that it would be possible to use `BHAMR<BinaryBHLevel::num_punctures>::particle_num_components ` but this results in circular includes. The only improvement I could come up with is storing `particle_num_components` elsewhere (e.g. in some extraction specific struct).
- I had to move the enum for `StateType` to a separate file, as I was again hitting circular includes. Was there a reason for defining it globally inside `GRAMRLevel`?


Source/GRTeclynCore/StateTypes.hpp
```diff
@@ -0,0 +1,17 @@
+/* GRTeclyn
+ * Copyright 2022 The GRTL collaboration.
+ * Please refer to LICENSE in GRTeclyn's root directory.
+ */
+
+#ifndef STATE_TYPES_HPP_
+#define STATE_TYPES_HPP_
+
+#pragma once
```
mirenradia: Let's stick to header guards (which are already here) rather than `#pragma once` for include protection.
```suggestion
```

Source/GRTeclynCore/SimulationParametersBase.hpp
```diff
@@ -15,7 +15,12 @@
 #include <limits>
 
 // add this type alias here for backwards compatibility
-using extraction_params_t = SphericalExtraction::params_t;
+static constexpr int extraction_num_components =
+    2; // As SphericalExtraction is templated over the number of components we
+       // need to provide the number here. Unfortunately, with the current
+       // structure we cannot avoid using magic numbers such as 2 here.
+using extraction_params_t = typename SphericalExtraction<
+    extraction_num_components>::params_t; // this is annoying!
```
mirenradia: I think I had a similar problem when I templated the `CCZ4RHS` class over the `deriv_t` type and I ended up just moving the `params_t` struct outside of it so that it's independent of that template parameter.

Source/GRTeclynCore/SimulationParametersBase.hpp
```diff
@@ -339,7 +344,7 @@ class SimulationParametersBase : public AMReXParameters
     CCZ4_params_t<> ccz4_params;
 
     bool activate_extraction{};
-    SphericalExtraction::params_t extraction_params;
+    SphericalExtraction<2>::params_t extraction_params;
```
mirenradia: Use `extraction_params_t` here to avoid using the magic number 2 again.

Source/ParticleInterpolator/CustomExtraction.hpp
```diff
@@ -0,0 +1,114 @@
...
+#include <string>
+#include <vector>
+
+// function to convert double to string with precision
+inline std::string to_string_with_precision(double value, int precision)
+{
+    std::ostringstream out;
+    out << std::fixed << std::setprecision(precision) << value;
+    return out.str();
+}
```
mirenradia: This seems like an odd place to put this. Is it used?

Source/ParticleInterpolator/LineExtraction.hpp
mirenradia, line 1: Perhaps this should be in the folder for a specific example where it is used but I'm happy to go with whatever @KAClough prefers.

Source/ParticleInterpolator/ParticleInterpolator.hpp
```diff
@@ -137,7 +138,9 @@ class ParticleInterpolator
                                  const amrex::Geometry &geom);
 
     // final interpolation routine exposed to the users
-    void interp(InterpolationQueryParticle &query,
+    // NEW: I need to carry through a_state_index here, as otherwise there are
+    // now out of scope errors
```
mirenradia: Can you just include the `StateTypes.hpp` header file instead or does that give you the errors?

Source/ParticleInterpolator/SurfaceExtraction.hpp
```diff
@@ -26,10 +27,12 @@
 #include <utility>
 #include <vector>
 
+template <int num_components> class ParticleInterpolator;
```
mirenradia: Is the forward declaration necessary if you already include the `ParticleInterpolator.hpp` header above?

Source/ParticleInterpolator/SurfaceExtraction.hpp
mirenradia: Let's rename these to `add_state_vars` and `add_derived_vars`.

mirenradia: They should also give an error if you try to:
1. add both state and derived vars (I think `InterpolationQueryParticle` should handle this).
2. add more vars than `num_components` (We should probably add this check somewhere in `ParticleInterpolator`, probably in `ParticleInterpolator::interp`).

Source/ParticleInterpolator/SurfaceExtraction.hpp
```diff
@@ -137,9 +140,13 @@ template <class SurfaceGeometry> class SurfaceExtraction
                       double a_restart_time = 0.0);
     // NOLINTEND(bugprone-easily-swappable-parameters)
 
-    //! Do the extraction
-    template <typename InterpAlgo>
-    void extract(AMRInterpolator<InterpAlgo> *a_interpolator);
+    //! Do the extraction for derived
+    void extract(ParticleInterpolator<num_components> *a_interpolator,
+                 amrex::Vector<BCParity> parities,
+                 const std::string &name_derived);
```
mirenradia: The logic for adding derived parities and passing the derived quantity name should probably be moved to `add_derived_vars` if possible.

Source/ParticleInterpolator/SurfaceExtraction.impl.hpp
```diff
@@ -122,64 +121,109 @@ void SurfaceExtraction<SurfaceGeometry>::add_diagnostic_vars(
...
     {
-        amrex::Abort("SurfaceExtraction: invalid AMRInterpolator pointer");
+        amrex::Abort("SurfaceExtraction: invalid ParticleInterpolator pointer");
     }
+
+    if (m_vars.size() != num_components)
+    {
+        amrex::Abort(
+            "SurfaceExtraction::extract: m_vars.size() != num_components");
+    }
```
mirenradia: Perhaps it should be possible to add fewer vars than `num_components` if you want to reuse a `SurfaceExtraction` object for two different sets of variables?

Source/Tagging/ExtractionTagger.hpp
```diff
@@ -31,7 +31,7 @@ class ExtractionTagger
     // Constructor takes only what it needs for extraction tagging
     // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
     ExtractionTagger(const double dx, const int a_level,
-                     const SphericalExtraction::params_t &a_params,
+                     const SphericalExtraction<2>::params_t &a_params,
```
mirenradia: This magic number 2 is definitely not ideal here!

Tests/SphericalExtractionTest/Make.package
```diff
@@ -0,0 +1,4 @@
+GRTECLYN_CEXE_headers += SphericalExtractionTest \
```
mirenradia:
```suggestion
GRTECLYN_CEXE_headers += SphericalExtractionTest.hpp \
```

Tests/SphericalExtractionTest/SphericalExtractionTest.cpp
```diff
@@ -0,0 +1,206 @@
...
+            if (amrex::ParallelDescriptor::MyProc() == 0)
+            {
+                // Trapezium rule should have second order convergence
+                CHECK(convergence_order_trapezium > 1.5);
+
+                // Simpson's rule should have fourth order convergence
+                CHECK(convergence_order_simpson > 3.5);
+
+                // Boole's rule should have sixth order convergence
+                CHECK(convergence_order_boole > 5.5);
```
mirenradia: No need to have the `INFO` statements if you use `CHECK_GT` instead of `CHECK`.

```suggestion
            if (amrex::ParallelDescriptor::MyProc() == 0)
            {
                // Trapezium rule should have second order convergence
                CHECK_GT(convergence_order_trapezium, 1.5);

                // Simpson's rule should have fourth order convergence
                CHECK_GT(convergence_order_simpson, 3.5);

                // Boole's rule should have sixth order convergence
                CHECK_GT(convergence_order_boole, 5.5);
```

Tests/SphericalExtractionTest/SphericalExtractionTest.cpp
```diff
@@ -0,0 +1,206 @@
...
+        ParticleInterpolator<2> interpolator;
+        interpolator.setup(&gr_amr, sim_params.boundary_params,
+                           sim_params.verbosity);
+
+        ParticleInterpolator<2> interpolator_hi;
+        interpolator_hi.setup(&gr_amr, sim_params.boundary_params,
+                              sim_params.verbosity);
+
+        // Low resolution spherical extraction
+        std::vector<int> evolution_vars = {c_phi, c_Pi};
```
mirenradia:
```suggestion
        std::vector<int> state_vars = {c_phi, c_Pi};
```

mirenradia (commented):
Thanks for the PR :) I have some minor comments but I think they should mostly be straightforward to fix.

Source/BlackHoles/BHAMR.hpp
```diff
@@ -25,9 +25,9 @@ template <int num_punctures> class BHAMR : public GRAMR
 
   public:
 
-    ParticleInterpolator<1>
-        m_weyl_interpolator; // weyl interpolator (used as chi interpolator in
-                             // this example)
+    static constexpr int particle_num_components = 2;
```
KAClough: Will this always just be used for Weyl interpolation? If so I would keep the "weyl" naming and make this `weyl_num_components` and then the interpolator `m_weyl_interpolator` as it was before. Then other interpolators can be created for other extractions, named in a similar way. A comment that we need to specify the number of components to be interpolated would probably be useful too.

Source/GRTeclynCore/StateTypes.hpp
```diff
@@ -0,0 +1,17 @@
...
+
+enum StateType
```
KAClough, line 14: Could we add a comment here to say what the state type is for? I always forget but I think it is basically just for the evolution state, and in principle we could have other types but we don't.

Source/ParticleInterpolator/LineExtraction.hpp
KAClough: I'm also not too bothered, but I would suggest that we call it LineExtraction and keep it here as a general tool. Maybe we could then have both the start and end points as inputs to make it a bit more general (and remove the length, so the direction is general).

Source/BlackHoles/BHAMR.hpp
```diff
@@ -41,6 +41,15 @@ template <int num_punctures> class BHAMR : public GRAMR
         }
     }
 
+    // set interpolator
+    void set_interpolator(
```
KAClough: Also consider changing naming here.

KAClough:
Hi, I added some very minor comments, so I think once you have addressed Miren's changes this is good to go! I do think we need a simpler version of the spherical extraction class, but we can either amend these classes later or maybe just add an additional class called `SimpleSphericalExtraction` that does a single scalar and integrates it. The main thing is that this seems to work, so we can use it to do some more testing of inspirals.

Source/ParticleInterpolator/SurfaceExtraction.impl.hpp
tamaraevst: Sounds good, I will change this check to when more vars are requested in extraction than available through interpolation.

Source/ParticleInterpolator/CustomExtraction.hpp
tamaraevst: Ah, this got inherited from: https://github.com/GRTLCollaboration/GRChombo/blob/main/Examples/ScalarFieldCosmo/CustomExtraction.hpp. I do not think we need this, so will remove these lines.

Source/GRTeclynCore/StateTypes.hpp
tamaraevst: Sounds good, I have added the comment the way I understand this.

Source/ParticleInterpolator/SurfaceExtraction.hpp
tamaraevst: Ok, I have changed the names. Unless I misunderstood something, addComp() in InterpolationQueryParticle should give an error when there is mixing in VariableType. When more vars are added than there are num_components, a check in extract() in SurfaceExtraction should catch this error.

Source/ParticleInterpolator/SurfaceExtraction.hpp
tamaraevst: I had a go at this. But it turned out quite convoluted with the current structure of the class, unless I have missed something obvious. I also got rid off the tuple that was storing variable indices, derivatives and so on and replaced it with a struct, where we now additionally store parities and the derived name.

Source/ParticleInterpolator/ParticleInterpolator.hpp
tamaraevst: Yes, this was my old work-around and I have forgotten to update it properly! Before I moved state type into a separate class.

Source/ParticleInterpolator/SurfaceExtraction.hpp
tamaraevst: Yeah, I do not think we need it here.

tamaraevst: Ah, no! I actually recalled why I have put it there. I needed it when compiling some tests. This was due to the wrong order of include files, which I have now fixed and, hence, removed the forward declaration in SurfaceExtraction.hpp.

tamaraevst: Sorry for the spam, but I do need to put this declaration back, because compilation order gets messed up otherwise. Changing include order gets overwritten by precommit anyways, and further we probably want the code to be independent of the include order.

tamaraevst:
Hi @mirenradia @KAClough I have addressed you comments. Please have a look and let me know. I replied directly to your comments in cases where it was not a straightforward fix. I have linter issues, but will resolve them once you are happy with the rest of the changes.

mirenradia:
> Hi @mirenradia @KAClough I have addressed you comments. Please have a look and let me know. I replied directly to your comments in cases where it was not a straightforward fix. I have linter issues, but will resolve them once you are happy with the rest of the changes.

@tamaraevst, I won't be able to re-review this until next week if that influences whether you want to start fixing the linter issues rather than waiting for me.

Source/ParticleInterpolator/SurfaceExtraction.hpp
mirenradia: I see the issue. This circular dependency is not great. Perhaps the cleanest solution is to move the `surface_extraction_params_t` and the derived `spherical_extraction_params_t` struct definitions to their own files. Then only the file for the latter would need to be included in `SimulationParameters.hpp`?

Source/ParticleInterpolator/SurfaceExtraction.hpp
mirenradia: Maybe it makes more sense for the error to appear when doing `add_var`?

Source/ParticleInterpolator/SurfaceExtraction.hpp
mirenradia: So quantities associated to a single derived name can have multiple components/vars e.g. `"Weyl4"` has 2 components (`Weyl4_Re` and `Weyl4_Im`). I don't think your code handles this case. Would you be able to adapt it to do so?

tamaraevst: Could you elaborate on what you mean, please? 

By `name_derived` in the `extract()` method I meant the derived group -- not the components. So `Weyl4` would be the group name, for example. Note that `name_derived` is used in [this instance](https://github.com/GRTLCollaboration/GRTeclyn/blob/feature/extraction/Source/ParticleInterpolator/SurfaceExtraction.impl.hpp#L213) of extraction, which is then passed onto the interpolating function, which later calls [derive](https://github.com/GRTLCollaboration/GRTeclyn/blob/develop/Source/ParticleInterpolator/ParticleInterpolator.impl.hpp#L341) for that group. So because we call derive on the group, both components will be computed, i.e. `Weyl4_Re` and `Weyl4_Im`. As long as `num_components = 2`, both of them should be extracted?

What does not work is if one were to use two different derived groups with the same ParticleInterpolator object and extraction method.

Source/ParticleInterpolator/SurfaceExtraction.hpp
tamaraevst: I guess you are referring to the second error in your list. If so, I have moved it to `add_var` now.

Source/ParticleInterpolator/SurfaceExtraction.hpp
mirenradia: OK, I think I understand now and I think only having one derived group is reasonable limitation. We can always change this in future if necessary.

Like for the number of components, can you move the error handling of multiple different derived quantities to `add_var` rather than `extract`?

P.S. When sharing links to specific lines of code, it's helpful to send a permalink to the specific commit rather than a branch which can change over time. You can get this by hitting `Y` on your keyboard before copying.

Examples/BinaryBH/BinaryBHLevel.cpp
mirenradia: Can you remove this code now that you have ported this capability?

Examples/BinaryBH/Main_BinaryBH.cpp
```diff
@@ -50,6 +50,12 @@ int runGRTeclyn(int /*argc*/, char * /*argv*/[])
 
     bh_amr.init(0., sim_params.stop_time);
 
+    ParticleInterpolator<
+        BHAMR<BinaryBHLevel::num_punctures>::weyl_num_components>
+        interpolator;
+    interpolator.setup(&bh_amr, sim_params.boundary_params, true);
+    bh_amr.set_weyl_interpolator(&interpolator);
```
mirenradia: Now that we are using this interpolator object just for interpolating $\Psi_4$, can we remove this code from here and the `BHAMR::set_weyl_interpolator` function. The interpolator instance can be stored in `BHAMR rather than just a pointer to it and the interpolator `setup` function can just be called in the `BHAMR` constructor (since you can access the parameter struct with `get_simulation_parameters()` member function). This will save other examples from having to duplicate this code.

We could add another layer of inheritance to `BHAMR` (maybe called something like `GWAMR`) which just has the interpolator object so that examples that want the Weyl interpolator but not the puncture tracker can use this but maybe it's worth deferring this for later/another PR.

Examples/BinaryBH/params_new.txt
mirenradia, line 1: Was this the parameter file you used for testing? I'm not sure we want to add this to the repository as we already have lots of parameter files and it's a bit confusing (hence #107). Maybe just add this as an attachment to a comment on this PR like you did in the first post.

Examples/BinaryBH/BinaryBHLevel.cpp
```diff
@@ -9,6 +9,7 @@
 #include "ChiTagger.hpp"
 #include "Constraints.hpp"
 #include "ExtractionTagger.hpp"
+#include "LineExtraction.hpp"
```
mirenradia: Is this used? If not, let's remove it.

Source/ParticleInterpolator/Make.package
```diff
@@ -7,7 +7,16 @@ GRTECLYN_CEXE_headers += ParticleInterpolator.hpp \
...
-                         DerivativeSetup.hpp
+                         DerivativeSetup.hpp \
+                         IntegrationMethod.hpp \
+                         IntegrationMethodSetup.hpp \
+                         SphericalExtraction.hpp \
+                         SphericalGeometry.hpp \
+                         SurfaceExtraction.hpp \
+                         SurfaceExtraction.impl.hpp \
+                         WeylExtraction.hpp \
+                         LineExtraction.hpp
```
mirenradia: Can you alphabetize this list?

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
```diff
@@ -39,7 +40,7 @@ void ParticleInterpolator<num_components>::setup(
 
     // read in the physical bounds for reflective BC checks (it is sufficient to
     // do this on lev = 0)
-    const amrex::Geometry &geom0 = m_gramr_ptr->getLevel(0).Geom();
+    const amrex::Geometry &geom0 = m_gramr_ptr->getLevel(state_index).Geom();
```
mirenradia: Even though `state_index == 0`, I think this is a bit misleading as really you want the label to be something like `coarsest_level = 0`

Source/ParticleInterpolator/SphericalExtraction.hpp
```diff
@@ -10,109 +10,81 @@
...
-        [[nodiscard]] const SurfaceExtraction::params_t &
-        get_surface_extraction_params() const
-        {
-            return *this;
-        }
-    };
+    using Base   = SurfaceExtraction<SphericalGeometry, num_components>;
+    using vars_t = typename Base::vars_t;
+    using mode_integrals_t =
+        std::pair<std::vector<double>, std::vector<double>>;
```
mirenradia, line 28: Maybe you can also add
```suggestion
using params_t = spherical_extraction_params_t;

```
and then replace uses of `spherical_extraction_params_t` with `params_t` below?

Source/ParticleInterpolator/SurfaceExtraction.hpp
```diff
@@ -9,65 +9,51 @@
...
+    using vars_t = var_t;
 
   protected:
     SurfaceGeometry m_geom; //!< the geometry class which knows about
                             //!< the particular surface
-    params_t m_params;
-    std::vector<std::tuple<int, VariableType, Derivative>>
-        m_vars; //!< the vector of pairs of
-    //!< variables and derivatives to extract
+    surface_extraction_params_t m_params;
```
mirenradia: Like my comment for `SphericalExtraction` can you add a
```cpp
using params_t = surface_extraction_params_t;
```
alias somewhere in this class? Hopefully they don't conflict with each other.

Source/ParticleInterpolator/SurfaceExtraction.hpp
```diff
@@ -9,65 +9,51 @@
...
-    };
+        int var{};
+        VariableType type{};
+        Derivative deriv;
 
-    using vars_t = std::tuple<int, VariableType, Derivative>;
+        amrex::Vector<BCParity> parities{};
+        std::string derived_name;
+    };
+    using vars_t = var_t;
```
mirenradia: Was there a reason you didn't just name the struct `vars_t`?

Source/ParticleInterpolator/SurfaceExtraction.impl.hpp
```diff
@@ -374,28 +423,29 @@ void SurfaceExtraction<SurfaceGeometry>::write_extraction(
...
                 {
-                    components[ivar] +=
-                        StateVariables::names[std::get<0>(m_vars[ivar])];
+                    components[ivar] += StateVariables::names[var.var];
                 }
                 else
                 {
-                    // components[ivar] +=
-                    //     DiagnosticVariables::names[std::get<0>(m_vars[ivar])];
+                    components[ivar] += std::to_string(var.var);
```
mirenradia: It would be better if it was possible to get the actual names of these derived variable components. It should be straightforward to do by accessing the `DeriveList` object (which you can get from `AmrLevel::get_derive_lst()`), getting the relevant `DeriveRec` using `DeriveList::get()` and then calling `variableName()` on this object.

Examples/BinaryBH/params_new.txt
tamaraevst: Ah, it migrated by accident, we do not need it here.

Examples/BinaryBH/Main_BinaryBH.cpp
tamaraevst: I made a change here although I needed to leave a designated setup() function for the interpolator inside `BHAMR`. I seem to not be able to move it directly into the constructor, as init() needs to be called and the level structure setup before the interpolator is called, otherwise I get segfaults.

mirenradia:
@tamaraevst, can you mark this PR as ready for review and rerequest a review from me if and when you'd like me to re-review?

tamaraevst:
> @tamaraevst, can you mark this PR as ready for review and rerequest a review from me if and when you'd like me to re-review?

I have finished the changes. One of the more non-trivial changes was in this comment [here](https://github.com/GRTLCollaboration/GRTeclyn/pull/174#discussion_r3314255743), unless I misunderstood your comment. I will not be looking at this until my return to the office, but otherwise I think it is ready, so I will mark it now.

Examples/BinaryBH/Main_BinaryBH.cpp
mirenradia: I see the issue. I think the cleanest solution is to:
1. Make `GRAMR::init()` `virtual` (as `amrex::Amr::init()` is)
2. Override `init` in `BHAMR` (as is done in `GRAMR::init()`) with the following body
   ```cpp
   {
       GRAMR::init(a_strt_time, a_stop_time);
   
       const auto &params = get_simulation_parameters();
       m_weyl_interpolator.setup(this, params.boundary_params, true);
   }
   ```
3. Remove the `BHAMR::set_weyl_interpolator()` function definition and call in Main_BinaryBH.cpp.

Hopefully that should work and it means that examples that use the `BHAMR` class don't need to call any separate setup functions.

Source/ParticleInterpolator/SurfaceExtraction.impl.hpp
```diff
@@ -442,7 +471,12 @@ void SurfaceExtraction<SurfaceGeometry, num_components>::write_extraction(
                 }
                 else
                 {
-                    components[ivar] += std::to_string(var.var);
+                    auto *derive_rec =
+                        amrex::AmrLevel::get_derive_lst().get(var.derived_name);
+                    //  amrex::Print() << "derived_name = " << var.derived_name
+                    // << ", var.var = " << var.var << ", variableName = " <<
+                    // derive_rec->variableName(var.var) << "\n";
```
mirenradia: Could you remove this print debugging code?

mirenradia:
I'm confused why the GitLab CI pipeline is failing. Any ideas, @julianakwan? All I can see in the output is
```
$ ${BINARY_DIR}/fcompare.gnu.ex ${ERR_TOL_FLAGS} ${OUTPUT_DIR}/BinaryBH_00008/ ${COMPARE_DIR}/BinaryBH/plt00008_compare
amrex::Error::0::Couldn't open file: /rds/project/rds-YVo7YUJF2mk/grteclyn-ci-svc/a100-test-1163770//BinaryBH_00008//Header !!!
SIGABRT
```
I thought we got over the error of files not appearing on the login node when we switched from `/home` (NFS) to `/rds` (Lustre). :thinking:

mirenradia:
I hacked in a fix for #185 in https://github.com/GRTLCollaboration/GRTeclyn/commit/e43a5084e8a6b542d5d2864eae2566ee2b306e81. It seems the Spherical Extraction Test is running out of memory. See [here](https://gitlab.developers.cam.ac.uk/rcs/rse/GRTeclyn/GRTeclyn/-/jobs/5751902) for the output.

tamaraevst:
Hi all, I have now resolved the two issues concerning the gitlab failing runs. 

First, I changed the CI for gitlab in the form that Miren suggested [here](https://github.com/GRTLCollaboration/GRTeclyn/tree/test/feature/extraction_gitlab_ci). 

Second, the gitlab run was failing when running the `params_test.txt` file with the BinaryBH example (where extraction was switched on). With the default extraction paramaters enforced in the file, there were some points being extracted very close to the axis of symmetry. Because we use `round` in `Langrange.hpp` to estimate the centering of the stencil, this pushed the stencil out of bounds. I therefore put a work-around for such a case [here](https://github.com/GRTLCollaboration/GRTeclyn/blob/9718c11fc5a6642e71d21ba413f9bedff80d5f3d/Source/ParticleInterpolator/Lagrange.hpp#L50).

I also thought that it would be appropriate to use e.g. `RoundOffLo()` from `Geometry` to check when the particle is off-domain (see [here](https://github.com/GRTLCollaboration/GRTeclyn/blob/9718c11fc5a6642e71d21ba413f9bedff80d5f3d/Source/ParticleInterpolator/ParticleInterpolator.impl.hpp#L44)). This change has also been implemeted in this PR.

KAClough:
Just to note that I tried running this on LUMI-G and with multiple GPU nodes I get a segfault (but not with a single node). Have you tried this @tamaraevst or @mirenradia ?

tamaraevst:
> Just to note that I tried running this on LUMI-G and with multiple GPU nodes I get a segfault (but not with a single node). Have you tried this @tamaraevst or @mirenradia ?

Nope, I only have access to 1 GPU. Do you have a backtrace/log file?

KAClough:
[Backtrace_32.txt](https://github.com/user-attachments/files/29217159/Backtrace_32.txt)

It only seems to happen when I use 8 nodes, maybe it is a LUMI problem specifically. This is the backtrace - I only get the ones for 32-63.

tamaraevst:
> [Backtrace_32.txt](https://github.com/user-attachments/files/29217159/Backtrace_32.txt)
> 
> It only seems to happen when I use 8 nodes, maybe it is a LUMI problem specifically. This is the backtrace - I only get the ones for 32-63.

Looks like this is coming from `populate_from_query()`. I wonder whether this has to do with `MFIter` at `lev=0` not existing on certain ranks.

Examples/BinaryBH/BinaryBHLevel.cpp
```diff
@@ -309,46 +310,30 @@ void BinaryBHLevel::specificPostTimeStep()
...
+
+            WeylExtraction my_extraction(simParams().extraction_params, m_dt,
+                                         m_time, first_step, restart_time);
+            my_extraction.execute_query(&get_bhamr_ptr()->m_weyl_interpolator);
         }
     }
 
+#if 0
+//xxxxx specificPostTimeStep
+    BL_PROFILE("BinaryBHLevel::specificPostTimeStep");
```
mirenradia: Can you move this profiler to the start of `BinaryBHLevel::specificPostTimeStep()`?

Source/ParticleInterpolator/Lagrange.hpp
```diff
@@ -26,15 +28,36 @@ template <int N> class Lagrange
...
+            10 * std::numeric_limits<
+                     amrex::Real>::epsilon(); // choose a small number around
+                                              // machine round-off precision
+        // when the position is very close to the axis of symmetry and
+        // reflective boundary conditions are used, center can end up being
+        // rounded up to -1. If center = -1, then the stencil is e.g. (-3, -2,
+        // -1, 0, 1) for 4th order interpolation, which is problematic here as
+        // -3 is out of bounds (note that we fill [4/2]=2 ghost cells). To avoid
+        // this, we will default to center = 0 in this situation.
+        if (lo_reflective &&
```
mirenradia: Why is `lo_reflective` needed here? Doesn't this apply whenever a particle is near the boundary (although I guess this is most likely in the `lo_reflective` case)? I guess it should probably be able to handle the case a particle is near the high boundary as well.

Source/ParticleInterpolator/SurfaceExtraction.impl.hpp
```diff
@@ -311,6 +390,7 @@ void SurfaceExtraction<SurfaceGeometry>::integrate()
     }
 
     // now broadcast result to non-zero ranks if requested
+    // T: this version does not work obviously. But we never broadcast anyways.
```
mirenradia: Why didn't you implement this? I think it would be reasonably straightforward with `amrex::ParallelDescriptor::Bcast`.

If you'd rather not bother, can you comment out the `a_broadcast_integral` arg in the signature so it isn't accidentally used? It can then be implemented later if desired.

Source/ParticleInterpolator/SurfaceExtraction.hpp
```diff
@@ -9,65 +9,51 @@
...
-        std::string data_path, integral_file_prefix;
-        std::string extraction_path, extraction_file_prefix;
-
-        int min_extraction_level()
-        {
-            return *(std::min_element(extraction_levels.begin(),
-                                      extraction_levels.end()));
-        }
-    };
+        int var{};
```
mirenradia: I wonder if it makes sense to rename this to `comp` or `icomp`? I find `var.var` (which appears in this file) confusing.

Tests/SphericalExtractionTest/SphericalExtractionTest.inputs
```diff
@@ -0,0 +1,47 @@
+#################################################
+# AMReX parameters
+
+amrex.the_arena_init_size = 0
```
mirenradia: Can you add a comment saying why you've put this here?

.gitlab-ci.yml
mirenradia, line 1: Can you revert the changes in this file?

mirenradia (changes requested):
I have a few minor comments that hopefully should be straightforward to fix.

Source/ParticleInterpolator/Lagrange.hpp
tamaraevst: Because it is the case with reflective BCs when one can have -1 as a center of the stencil. The general case when a point lies outside the physical boundary is handled in check_domain.

Source/ParticleInterpolator/SurfaceExtraction.impl.hpp
tamaraevst: I left it as it was initially in the file. I do not think we use broadcast anywhere.

Source/ParticleInterpolator/Lagrange.hpp
mirenradia: Because you're now using the roundoff limits in `check_domain`?

Source/ParticleInterpolator/SurfaceExtraction.impl.hpp
mirenradia: I think I added this when I implemented these classes in GRChombo in order to retain backwards compatibility with how `WeylExtraction` previously behaved. There may be a use case but I agree we don't need it for now. 

I'm now wondering if it's worth just leaving this codepath in but putting an `amrex::Error` here saying that this hasn't been implemented.

Source/ParticleInterpolator/Lagrange.hpp
tamaraevst: Ah I think I misread your comment. I think it may also have an issue if one queries the point right at the physical boundary. But we can use change `x[d] < lo_g `to` x[d] <= lo_g` etc. in check_domain as an easy fix (to not allow points right on physical boundary), what do you think?

Source/ParticleInterpolator/SurfaceExtraction.impl.hpp
tamaraevst: Ok, I will put the error whenever `a_broadcast_integral` is asked to be `true`.

tamaraevst:
Hi @mirenradia @KAClough I have tidied up this PR based off latest Miren's comments. I added the handling of interpolation when the user request a point at the high boundary which also has symmetric BC imposed (see [here](https://github.com/GRTLCollaboration/GRTeclyn/blob/edc260740169aa0d825934975959b90c927e8dbf/Source/ParticleInterpolator/Lagrange.hpp#L64)). It turned out to be a bit more ugly than the case with the low boundary and symmetric BC (because of how we determine the `grid_pos` e.g. [here](https://github.com/GRTLCollaboration/GRTeclyn/blob/edc260740169aa0d825934975959b90c927e8dbf/Source/ParticleInterpolator/Lagrange.hpp#L127). 

The gitlab CI is failing though, looks like maybe an issue on the CSD3 end?

Source/ParticleInterpolator/Lagrange.hpp
```diff
@@ -42,16 +42,29 @@ template <int N> class Lagrange
...
+        // 4th order interpolation, which is problematic here as -3 is out of
+        // bounds (note that we fill [4/2]=2 ghost cells). To avoid this, we
+        // will default to center = 0 in this situation.
+
+        const auto lo_face = amrex::Real(
+            -0.5); // for low symmetric boundary, the face is at -0.5
+        const auto hi_face =
+            amrex::Real(ncell) -
+            amrex::Real(
+                0.5); // for high symmetric boundary, the face is at ncell-0.5
```
mirenradia, line 58: I thought it was at `ncell + 0.5`?

mirenradia: Ok, now I seen you have used the length rather than the hi indices of the box so I think it's correct assuming that the lo indices are at $(0,0,0)$ but perhaps a bit inconsistent. This is something that will need to be changed for #149 but I'm happy to leave it for now.

mirenradia (approved):
The CSD3 workflow is failing because CSD3 is down at the moment but I have manually run the unit tests and regressions test on the AMD MI300A node in COSMA and everything passes.

I'm happy for this to be merged.

PR #175, opened 2026-04-15
CI: Add dependabot-style workflow for doctest

mirenradia:
[doctest](https://github.com/doctest/doctest) seems to now be maintained and receiving regular updates. Unfortunately, since we include the header file directly in this repo (instead of e.g. using git submodules), we can't rely on dependabot to check for updates hence the addition of this CI workflow.

### DIFF
--- a/.github/workflows/update-doctest.yml
+++ b/.github/workflows/update-doctest.yml
@@ -1,0 +1,72 @@
+name: Update doctest
+
+on:
+    schedule:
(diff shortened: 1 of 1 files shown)


mirenradia:
I was going to manually update doctest to the latest version (2.5.2) but I thought it would be best to get this workflow to do it to check that it works.

julianakwan (approved):
Thanks for the PR! It looks good to me and I was able to double check everything up to making the actual PR with `act`. I guess we will know next Monday!

PR #176, opened 2026-04-28
Bump to C++20

julianakwan:
This will bump up the required version of C++ from C++17 to C++20 in keeping with the recent changes to AMReX in AMReX-Codes/amrex#5265 and fix the failing CI/CD pipelines. 

It will also increase the versions of CUDA, ROCm, and oneAPI required for the CI in the GPU build and the version of CUDA and GCC in the GitLab regression tests run on the Ampere nodes.


.github/workflows/gpu-build.yml
mirenradia, line 1: Could we use more recent versions for all of the software stacks? By default, I would use the latest (e.g. ROCm 7.2.2) but for CUDA, I wouldn't object to the last in the 12.x series (12.9.1). For oneAPI, the 2026.0 release came out today!

mirenradia (changes requested):
Thanks for the PR. Just one small comment for the GPU Build workflow.

.github/workflows/gpu-build.yml
julianakwan: Done!

mirenradia (approved):
LGTM

PR #186, opened 2026-06-09
CI: Improve GitLab Pipeline

### LINKED ISSUE #185: CI: GitLab pipeline no longer prints output during run
I noticed in https://github.com/GRTLCollaboration/GRTeclyn/pull/174#issuecomment-4577047992 that the GitLab CI pipeline no longer prints any output when the `run-examples.sh` job is run. 

Definitely my fault as I should have picked this up when reviewing #121.

mirenradia:
This PR makes two improvements to the GitLab pipeline:

* Makes the test job exit when it encounters its first error.
* Prints the output of the test job that is submitted as a SLURM job (before exiting if it fails). This resolves #185.

An example of the csd3-a100-test job failing can be seen [here](https://gitlab.developers.cam.ac.uk/rcs/rse/GRTeclyn/GRTeclyn/-/jobs/5766393).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -53,7 +53,24 @@ csd3-a100-test:
       if [[ ! -d ${OUTPUT_DIR} ]]; then
         mkdir ${OUTPUT_DIR}
       fi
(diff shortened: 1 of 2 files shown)


julianakwan (approved):
Thanks for fixing the CI pipeline! I guess we didn't notice it in #121 because the CI didn't fail 😆

PR #189, opened 2026-07-08
CI: Make lint-ignore workflow less annoying

mirenradia:
Allow some patterns to be ignored by the linter but not trigger a comment on GitHub PRs. These are patterns prefixed with `lint-only:`. Add this prefix to the `*.impl.hpp` pattern.

### DIFF
--- a/.lint-ignore
+++ b/.lint-ignore
@@ -2,9 +2,11 @@
 # GitHub action. These are mostly files that have yet to be ported to AMReX
 # Once these files have been ported to AMReX, they should be removed from this
 # file so they are linted.
+# Prefix a pattern with "lint-only:" to ignore it in linting without
+# triggering the .lint-ignore PR reminder workflow.
 
 # These files don't make sense on their own
-*.impl.hpp
+lint-only: *.impl.hpp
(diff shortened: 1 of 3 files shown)


.lint-ignore
```diff
@@ -2,9 +2,11 @@
 # GitHub action. These are mostly files that have yet to be ported to AMReX
 # Once these files have been ported to AMReX, they should be removed from this
 # file so they are linted.
+# Prefix a pattern with "lint-only:" to ignore it in linting without
+# triggering the .lint-ignore PR reminder workflow.
 
 # These files don't make sense on their own
-*.impl.hpp
+lint-only: *.impl.hpp
 Tests/CCZ4GeometryUnitTests/CCZ4GeometryMathematicaValues.hpp
```
julianakwan: Can you please update this file with CCZ4GeometryExpectedValues.hpp? Thanks!

julianakwan (changes requested):
Thanks for this PR! Looks good, just one minor change. 

To be perfectly honest, I never found lint-ignore annoying just the linter itself.

.lint-ignore
mirenradia: Done

julianakwan (approved):
Thanks for the change! Looks good to merge!

PR #190, opened 2026-07-08
Rename example executable prefixes

mirenradia:
A very small PR to rename our example executable prefixes from the default `main` to their example name e.g. `BinaryBH` or `KleinGordon`. Now instead of executables being something like

```
main3d.gnu.MPI.ex
```

they'll instead be be

```
BinaryBH3d.gnu.MPI.ex
```

I think one of @julianakwan or I did this before in another PR but it was never merged for some reason.

The main motivation for doing this now is that I want to modify the prefix in the case of TwoPunctures (#18).

### DIFF
--- a/.github/workflows/regression-tests-examples-template.yml
+++ b/.github/workflows/regression-tests-examples-template.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-24.04
     env:
       AMREX_HOME: ${{ github.workspace }}/amrex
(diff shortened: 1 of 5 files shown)


KAClough (approved):
LGTM

mirenradia:
Can we hold off merging this until CSD3 is back (hopefully today) just to check I haven't broken the GitLab CI pipeline.

I think it would be best to merge all the commits and *not* squash them.

mirenradia:
The GitLab CI has passed so I will merge this PR.

PR #193, opened 2026-07-20
Add derivative support to ParticleInterpolator

### LINKED ISSUE #143: Make interpolation compatible with derivatives
In #141 we add support for particle interpolation of state variables and no derivatives. For apparent horizon and CCE, it is desirable to be able to interpolate the derivatives as well. We will need to adjust weights in the Lagrange interpolation.

AdamG412:
Add the ability to interpolate first and second derivatives using the ParticleInterpolator. Works by taking the derivative of the function to generate Lagrange weights used by amrex:

```math
w_j(x) = \prod_{i \neq j} \frac{x - x_i}{x_j - x_i}
```
```math
w_j^{\prime}(x) = \sum_{k \neq j} \frac{1}{x_j - x_k} \prod_{i \neq j, i \neq k} \frac{x - x_i}{x_j - x_i}
```
```math
w_j^{\prime\prime}(x) = \sum_{l \neq j} \sum_{k \neq j, k \neq l} \frac{1}{(x_j - x_k)(x_j - x_l)} \prod_{i \neq j, i \neq k, i \neq l} \frac{x - x_i}{x_j - x_i}
```

And selecting the set of weights corresponding to the required derivative in each dimension.
Tests the interpolation against polynomials of lower order than our interpolation, so we should be accurate up to machine precision. 
The changes needed to work with the different derivatives should also fix the problem of previously only being able to interpolate variables with contiguous comp numbers, as we now loop over our comps (moved into arrays from the orignal `derivative` -> `comps_t` map to be able to run on gpu) and get the variable index directly from each `out_t` structure.

Closes #143

### DIFF
--- a/Source/ParticleInterpolator/Lagrange.hpp
+++ b/Source/ParticleInterpolator/Lagrange.hpp
@@ -266,9 +266,11 @@ template <int N> class Lagrange
     // NOLINTBEGIN(bugprone-easily-swappable-parameters)
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
     interpolate(const amrex::Array4<amrex::Real const> *data_arr,
-                amrex::ParticleReal *val, const Derivative *derivs,
-                InterpolationQueryParticle::out_t *const *comps_arr,
-                const int *comp_counts, int n_deriv,
+                amrex::ParticleReal *val,
+                const amrex::Gpu::ManagedVector<Derivative> derivs,
+                amrex::Gpu::ManagedVector<InterpolationQueryParticle::out_t *>
+                    comps_ptr,
+                amrex::Gpu::ManagedVector<int> comp_counts, int n_deriv,
                 amrex::Real const dxi) const
     // NOLINTEND(bugprone-easily-swappable-parameters)
     {
@@ -283,7 +285,7 @@ template <int N> class Lagrange
         {
             const Derivative &deriv = derivs[i];
 
-            const InterpolationQueryParticle::out_t *comps = comps_arr[i];
+            const InterpolationQueryParticle::out_t *comps = comps_ptr[i];
 
             // Collect dimensions where we want at least a first derivative
             for (int dim = 0; dim < AMREX_SPACEDIM; ++dim)
--- a/Source/ParticleInterpolator/Lagrange.hpp
+++ b/Source/ParticleInterpolator/Lagrange.hpp
@@ -236,12 +236,12 @@ template <int N> class Lagrange
                   (amrex::Real(par.pos(2)) - plo[2]) * dxi[2] -
                   static_cast<amrex::Real>(!is_nodal[2]) * amrex::Real(0.5););
 
-        build_stencil(xpos, i0, weights_local[0], lo_reflective[1],
-                      hi_reflective[1], domain_ncell[1], 0);
-        build_stencil(xpos, i0, weights_d1[0], lo_reflective[1],
-                      hi_reflective[1], domain_ncell[1], 1);
-        build_stencil(xpos, i0, weights_d2[0], lo_reflective[1],
-                      hi_reflective[1], domain_ncell[1], 2);
+        build_stencil(xpos, i0, weights_local[0], lo_reflective[0],
+                      hi_reflective[0], domain_ncell[1], 0);
+        build_stencil(xpos, i0, weights_d1[0], lo_reflective[0],
+                      hi_reflective[0], domain_ncell[0], 1);
+        build_stencil(xpos, i0, weights_d2[0], lo_reflective[0],
+                      hi_reflective[0], domain_ncell[0], 2);
 #if AMREX_SPACEDIM >= 2
         build_stencil(ypos, j0, weights_local[1], lo_reflective[1],
                       hi_reflective[1], domain_ncell[1], 0);
@@ -253,12 +253,12 @@ template <int N> class Lagrange
 #endif
 
 #if AMREX_SPACEDIM == 3
-        build_stencil(zpos, k0, weights_local[2], lo_reflective[1],
-                      hi_reflective[1], domain_ncell[1], 0);
-        build_stencil(zpos, k0, weights_d1[2], lo_reflective[1],
-                      hi_reflective[1], domain_ncell[1], 1);
-        build_stencil(zpos, k0, weights_d2[2], lo_reflective[1],
-                      hi_reflective[1], domain_ncell[1], 2);
+        build_stencil(zpos, k0, weights_local[2], lo_reflective[2],
+                      hi_reflective[2], domain_ncell[2], 0);
+        build_stencil(zpos, k0, weights_d1[2], lo_reflective[2],
+                      hi_reflective[2], domain_ncell[2], 1);
+        build_stencil(zpos, k0, weights_d2[2], lo_reflective[2],
+                      hi_reflective[2], domain_ncell[2], 2);
 #endif
     }
 
--- a/Source/ParticleInterpolator/Lagrange.hpp
+++ b/Source/ParticleInterpolator/Lagrange.hpp
@@ -266,11 +266,9 @@ template <int N> class Lagrange
     // NOLINTBEGIN(bugprone-easily-swappable-parameters)
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
     interpolate(const amrex::Array4<amrex::Real const> *data_arr,
-                amrex::ParticleReal *val,
-                const amrex::Gpu::ManagedVector<Derivative> derivs,
-                amrex::Gpu::ManagedVector<InterpolationQueryParticle::out_t *>
-                    comps_ptr,
-                amrex::Gpu::ManagedVector<int> comp_counts, int n_deriv,
+                amrex::ParticleReal *val, const Derivative *derivs,
+                InterpolationQueryParticle::out_t *const *comps_arr,
+                const int *comp_counts, int n_deriv,
                 amrex::Real const dxi) const
     // NOLINTEND(bugprone-easily-swappable-parameters)
     {
@@ -285,7 +283,7 @@ template <int N> class Lagrange
         {
             const Derivative &deriv = derivs[i];
 
-            const InterpolationQueryParticle::out_t *comps = comps_ptr[i];
+            const InterpolationQueryParticle::out_t *comps = comps_arr[i];
 
             // Collect dimensions where we want at least a first derivative
             for (int dim = 0; dim < AMREX_SPACEDIM; ++dim)
--- a/Source/ParticleInterpolator/Lagrange.hpp
+++ b/Source/ParticleInterpolator/Lagrange.hpp
@@ -98,10 +98,10 @@ template <int N> class Lagrange
         switch (deriv_order)
         {
         case 1:
-            poly_interp_coeff_d1(0.0, rel_pos, N, weights);
+            poly_interp_coeff_d1(0.0, rel_pos, weights);
             break;
         case 2:
-            poly_interp_coeff_d2(0.0, rel_pos, N, weights);
+            poly_interp_coeff_d2(0.0, rel_pos, weights);
             break;
         default:
             amrex::poly_interp_coeff(0.0, rel_pos, N, weights);
@@ -114,13 +114,12 @@ template <int N> class Lagrange
 
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void
     poly_interp_coeff_d1(amrex::Real xInt, amrex::Real const *AMREX_RESTRICT x,
-                         int order,
                          amrex::Real *AMREX_RESTRICT weights) noexcept
     {
-        for (int j = 0; j < order; ++j)
+        for (int j = 0; j < N; ++j)
         {
             auto den = amrex::Real(1.0);
-            for (int i = 0; i < order; ++i)
+            for (int i = 0; i < N; ++i)
             {
                 if (i != j)
                 {
@@ -130,14 +129,14 @@ template <int N> class Lagrange
 
             weights[j] = amrex::Real(0.0);
 
-            for (int k = 0; k < order; ++k)
+            for (int k = 0; k < N; ++k)
             {
                 if (k == j)
                 {
                     continue;
                 }
                 auto num = amrex::Real(1.0);
-                for (int i = 0; i < order; ++i)
+                for (int i = 0; i < N; ++i)
                 {
                     if (i != j && i != k)
                     {
@@ -152,15 +151,14 @@ template <int N> class Lagrange
 
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void
     poly_interp_coeff_d2(amrex::Real xInt, amrex::Real const *AMREX_RESTRICT x,
-                         int order,
                          amrex::Real *AMREX_RESTRICT weights) noexcept
     {
-        for (int j = 0; j < order; ++j)
+        for (int j = 0; j < N; ++j)
         {
             weights[j] = amrex::Real(0.0);
 
             auto den = amrex::Real(1.0);
-            for (int i = 0; i < order; ++i)
+            for (int i = 0; i < N; ++i)
             {
                 if (i != j)
                 {
@@ -168,14 +166,14 @@ template <int N> class Lagrange
                 }
             }
 
-            for (int l = 0; l < order; ++l)
+            for (int l = 0; l < N; ++l)
             {
                 if (l == j)
                 {
                     continue;
                 }
 
-                for (int k = 0; k < order; ++k)
+                for (int k = 0; k < N; ++k)
                 {
                     if (k == j)
                     {
@@ -188,7 +186,7 @@ template <int N> class Lagrange
 
                     auto num = amrex::Real(1.0);
 
-                    for (int i = 0; i < order; ++i)
+                    for (int i = 0; i < N; ++i)
                     {
                         if (i != j && i != k && i != l)
                         {
@@ -239,7 +237,7 @@ template <int N> class Lagrange
                   static_cast<amrex::Real>(!is_nodal[2]) * amrex::Real(0.5););
 
         build_stencil(xpos, i0, weights_local[0], lo_reflective[0],
-                      hi_reflective[0], domain_ncell[1], 0);
+                      hi_reflective[0], domain_ncell[0], 0);
         if (need_d1[0])
         {
             build_stencil(xpos, i0, weights_d1[0], lo_reflective[0],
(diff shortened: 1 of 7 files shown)


Source/ParticleInterpolator/Lagrange.hpp
```diff
@@ -128,57 +236,112 @@ template <int N> class Lagrange
...
 
 #if AMREX_SPACEDIM == 3
-        build_stencil(zpos, k0, weights_z, lo_reflective[2], hi_reflective[2],
-                      domain_ncell[2]);
+        build_stencil(zpos, k0, weights_local[2], lo_reflective[2],
+                      hi_reflective[2], domain_ncell[2], 0);
+        build_stencil(zpos, k0, weights_d1[2], lo_reflective[2],
+                      hi_reflective[2], domain_ncell[2], 1);
+        build_stencil(zpos, k0, weights_d2[2], lo_reflective[2],
+                      hi_reflective[2], domain_ncell[2], 2);
```
mirenradia: Would it be possible to make it only call `build_stencil` for the derivative orders it actually needs.

Most of the time, we won't need the derivative stencils and it seems wasteful to calculate it every time.

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
```diff
@@ -286,7 +298,8 @@ void ParticleInterpolator<num_components>::interpolate_to_particle(
 
                 amrex::ParticleReal interpolated_vals[ncomp];
                 lagrange_interp.interpolate(&fab_array, interpolated_vals,
-                                            start_comp, ncomp);
+                                            derivs, comps, comp_counts,
+                                            num_derivs, 1 / dxi[0]);
```
mirenradia: Rather than pass `1 / dxi[0]` and then divide by it in `interpolate()`, you might as well pass `dxi[0]` directly and then multiply by it.

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
```diff
@@ -270,6 +267,21 @@ void ParticleInterpolator<num_components>::interpolate_to_particle(
         const int num_particles = par_iter.numParticles();
         auto fab_array          = mfab[par_iter].const_array();
 
+        Derivative derivs[ncomp];
+        InterpolationQueryParticle::out_t *comps[ncomp];
+        int comp_counts[ncomp];
```
mirenradia: Just checking these arrays are the right lengths? They look like they're `ncomp` here but then below they're indexed into with `num_derivs`?

Tests/ParticleInterpolatorUnitTest/ParticleInterpolatorUnitTest.cpp
```diff
@@ -166,9 +185,30 @@ void run_particle_interpolator_test()
...
+                 << " z = " << z << ". The true value should be "
+                 << B_known_dxdx);
+
             CHECK(A_local[ipoint] == doctest::Approx(A_known).epsilon(1e-10));
+            CHECK(A_dx[ipoint] == doctest::Approx(A_known_dx).epsilon(1e-10));
+            CHECK(A_dxdy[ipoint] ==
+                  doctest::Approx(A_known_dxdy).epsilon(1e-10));
             CHECK(B_local[ipoint] == doctest::Approx(B_known).epsilon(1e-10));
+            CHECK(B_dxdx[ipoint] ==
+                  doctest::Approx(B_known_dxdx).epsilon(1e-10));
```
mirenradia: Could you replace these `INFO`s and `CHECK`s and `CHECK_MESSAGE`s so that only the relevant info is printed for each check. See the [doctest docs](https://github.com/doctest/doctest/blob/master/doc/markdown/assertions.md) for more info.

mirenradia (commented):
Thanks @AdamG412 for the PR.

I'll let @tamaraevst review it properly but I have a couple of small comments.

mirenradia:
I slightly modified the formatting of your PR description to make the maths render in displayed mode.

tamaraevst:
Hi @AdamG412 thanks for this PR. Could you please let me know once it is ready and you want me to review this (you should be able to request a review from me on this page)?

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
```diff
@@ -267,24 +267,28 @@ void ParticleInterpolator<num_components>::interpolate_to_particle(
...
+            derivs[i]      = comps_it->first;
+            comps[i]       = comps_it->second.data();
+            comp_counts[i] = static_cast<int>(comps_it->second.size());
+            ++i;
         }
 
         amrex::ParallelFor(
             num_particles,
-            [=] AMREX_GPU_DEVICE(int ip)
+            [&] AMREX_GPU_DEVICE(int ip)
```
mirenradia: I think GPU Lambdas always need to capture by value.

mirenradia: In fact this is exactly where the GPU Build CI workflow seems to be failing.

AdamG412: Ah yes, thought I'd fixed this already

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
```diff
@@ -270,6 +297,42 @@ void ParticleInterpolator<num_components>::interpolate_to_particle(
         const int num_particles = par_iter.numParticles();
         auto fab_array          = mfab[par_iter].const_array();
 
+        // Gather comp map into arrays so it can be used on GPU
+        int num_derivs =
```
tamaraevst: Is this not what you have done just before entering the `ParIterType` loop? Do you need this repetition here again?

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
```diff
@@ -270,6 +297,42 @@ void ParticleInterpolator<num_components>::interpolate_to_particle(
...
+        for (auto comps_it = m_query->compsBegin();
+             comps_it != m_query->compsEnd(); ++comps_it)
+        {
+            derivs[i]      = comps_it->first;
+            comps[i]       = comps_it->second.data();
+            comp_counts[i] = static_cast<int>(comps_it->second.size());
+            ++i;
+        }
+
+        amrex::GpuArray<bool, AMREX_SPACEDIM> need_d1{};
```
tamaraevst: Can the filling of `need_d1` and `need_d2` be moved outside the loop? (e.g. where you gather comp map into managed arrays)

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
```diff
@@ -262,6 +259,36 @@ void ParticleInterpolator<num_components>::interpolate_to_particle(
...
+    // Gather comp map into managed arrays
+    const int num_derivs = static_cast<int>(
+        std::distance(m_query->compsBegin(), m_query->compsEnd()));
+
+    amrex::Gpu::ManagedVector<Derivative> derivs(num_derivs);
+    amrex::Gpu::ManagedVector<int> comp_counts(num_derivs);
+
+    // Flatten comps and generate pointers into it
+    amrex::Gpu::ManagedVector<InterpolationQueryParticle::out_t> comps_flat(
+        ncomp);
```
tamaraevst, line 267: Here we use `const int ncomp = num_components` to define `comps_flat`. But we also have `m_query->numComps()`. Shall we add a check that the two are the same via `AMREX_ASSERT`?

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
tamaraevst: I think this assumes contiguous order, so it will not work here anymore. Or am I missing something?

Source/ParticleInterpolator/Lagrange.hpp
```diff
@@ -87,33 +94,136 @@ template <int N> class Lagrange
...
         // Write in the base (starting) index
         base_idx = stencil[0];
     }
 
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void
+    poly_interp_coeff_d1(amrex::Real xInt, amrex::Real const *AMREX_RESTRICT x,
+                         int order,
+                         amrex::Real *AMREX_RESTRICT weights) noexcept
+    {
+        for (int j = 0; j < order; ++j)
```
tamaraevst: `order` here is the number of points in the stencil, so can we not use `N` here directly?

Source/ParticleInterpolator/Lagrange.hpp
```diff
@@ -128,57 +238,131 @@ template <int N> class Lagrange
...
-#if AMREX_SPACEDIM >= 2
                 }
+                else if (deriv[dim] == 2)
+                {
+                    for (int j = 0; j < N; j++)
+                    {
+                        weights[dim][j] = weights_d2[dim][j] * pow(dxi, 2);
+                    }
+                }
+                else
```
tamaraevst, line 335: I wonder whether this should throw an error? We do not support derivatives past 2nd order.

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
```diff
@@ -658,7 +724,7 @@ void ParticleInterpolator<num_components>::apply_parity_and_store_values()
             const int comp           = entry.comp;
             amrex::ParticleReal *out = entry.out_data_ptr;
 
-            const int k = comp - start_comp; // reindex from 0
+            const int k = comp - start_comp;
```
tamaraevst: Do we need `start_comp` and `k` here? I think the assert can also fail now since we do not have contiguous order.

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
tamaraevst, line 231: I think an important distinction between the previous implementation and introducing derivatives is that any added derivative becomes a new component. So, it should be made clear to the user that, for example, if they interpolate $\chi$, and $h_{11}$ and $\partial_x \chi$, they are requesting 3 components, not 2. Can we add a comment on this in the `ParticleInterpolator` class? Sorry to be pedantic here... This of course should be somewhat obvious, since we use `num_components` to size the particle SoA. But it will not hurt noting this, especially for those that will not be looking into the source code. I will add this into the docs as well, once we merge this PR.

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
```diff
@@ -262,6 +259,36 @@ void ParticleInterpolator<num_components>::interpolate_to_particle(
...
+        ncomp);
+    amrex::Gpu::ManagedVector<InterpolationQueryParticle::out_t *> comps_ptr(
+        num_derivs);
+
+    int i = 0, off = 0;
+    for (auto comps_it = m_query->compsBegin(); comps_it != m_query->compsEnd();
+         ++comps_it, ++i)
+    {
+        derivs[i]      = comps_it->first;
+        comp_counts[i] = static_cast<int>(comps_it->second.size());
```
tamaraevst, line 281: So `comp_counts` keeps track of the size of the derivative group? E.g. if we have:

LOCAL: [chi, K, A_11]
DX: [chi]

Then, `comp_counts[0] = 3` and `comp_counts[1] = 1`. Can we add a comment, please?

Source/ParticleInterpolator/Lagrange.hpp
```diff
@@ -128,57 +238,131 @@ template <int N> class Lagrange
...
-                    for (int ii = 0; ii < N; ++ii)
+                    for (int j = 0; j < N; j++)
                     {
-                        val[counter] +=
-                            data(amrex::IntVect(
-                                     AMREX_D_DECL(i0 + ii, j0 + jj, k0 + kk)),
-                                 comp) *
-                            AMREX_D_TERM(weights_x[ii], *weights_y[jj],
-                                         *weights_z[kk]);
+                        weights[dim][j] = weights_d1[dim][j] * dxi;
```
tamaraevst: Can we add a comment here that `dxi` factor is required here through the chain rule by virtue of how we compute `xpos`, `ypos` etc.?

Source/ParticleInterpolator/Lagrange.hpp
```diff
@@ -128,57 +238,131 @@ template <int N> class Lagrange
...
 
     // Function to perform the interpolation
+    // NOLINTBEGIN(bugprone-easily-swappable-parameters)
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
     interpolate(const amrex::Array4<amrex::Real const> *data_arr,
-                amrex::ParticleReal *val, int start_comp, int ncomp) const
+                amrex::ParticleReal *val, const Derivative *derivs,
+                InterpolationQueryParticle::out_t *const *comps_arr,
+                const int *comp_counts, int n_deriv,
+                amrex::Real const dxi) const
```
tamaraevst: Shall we pass the vector of `dxi` for each dimension? Like in `compute_weights`?

Source/ParticleInterpolator/Lagrange.hpp
```diff
@@ -128,57 +238,131 @@ template <int N> class Lagrange
                   (amrex::Real(par.pos(2)) - plo[2]) * dxi[2] -
                   static_cast<amrex::Real>(!is_nodal[2]) * amrex::Real(0.5););
 
-        build_stencil(xpos, i0, weights_x, lo_reflective[0], hi_reflective[0],
-                      domain_ncell[0]);
-
+        build_stencil(xpos, i0, weights_local[0], lo_reflective[0],
+                      hi_reflective[0], domain_ncell[1], 0);
```
tamaraevst: `domain_ncell[1]` looks like a typo?

Tests/ParticleInterpolatorUnitTest/PolynomialDerivedQuantity.hpp
```diff
@@ -113,5 +116,6 @@ class PolynomialDerivedQuantity
   private:
     int m_c_poly; // destination comp
 };
+// NOLINTEND(bugprone-easily-swappable-parameters)
```
tamaraevst: I wonder whether this lint end-line can be pushed to a bit earlier on?

Tests/ParticleInterpolatorUnitTest/ParticleInterpolatorUnitTest.cpp
```diff
@@ -156,19 +171,42 @@ void run_particle_interpolator_test()
...
-                 << " z = " << z << ". The true value should be " << A_known);
-            INFO("Interpolated B is "
-                 << B_local[ipoint] << " at point x = " << x << " y = " << y
-                 << " z = " << z << ". The true value should be " << B_known);
-
-            CHECK(A_local[ipoint] == doctest::Approx(A_known).epsilon(1e-10));
-            CHECK(B_local[ipoint] == doctest::Approx(B_known).epsilon(1e-10));
+            double A_known      = pow(x, 3) * pow(y, 3) * pow(z, 3);
+            double A_known_dx   = 3 * pow(x, 2) * pow(y, 3) * pow(z, 3);
+            double A_known_dxdy = 3 * pow(x, 2) * 3 * pow(y, 2) * pow(z, 3);
```
tamaraevst: Maybe here it is worth taking the derivative w.r.t. $z$ for one of the cases so that we test parity application when symmetric BCs are used?

tamaraevst (changes requested):
Hi @AdamG412, thanks for this PR. I attach my preliminary comments, but will re-review once more changes are made. 

On the side note, I made a minor stylistic change in your PR text of $w(j) \rightarrow w_j(x)$ to make the notation a little clearer. 

I like the idea of having the non-contiguous order of components in the `ParticleInterpolator`. But we need additional changes to make it work properly (see my comments above). I wonder whether we should explicitly test it, using e.g. `LineExtraction`. Either yourself or myself can do it at a later stage once everything is in working order (and this is probably to be also discussed with @KAClough and @mirenradia). 

Also note that there are changes incoming in PR #211, so you may want to sync in your branch with the changes made there.

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
AdamG412: No, looks like I just forgot to remove it when I reworked this section

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
AdamG412: Yes I think you're right, I didn't look at this before. I've also run into the issue in the AHFinder that if you have too many comps in your query (e.g. if you had local and first derivatives for all grid variables) this crashes because ncomp becomes larger than the number of actual components. I also think that in general if you want to interpolate a comp with more than one derivative you'll be filling the ghost cells with unneeded data.
The contiguous order I'm not sure we can get around, but I think we could get around the second issue by tracking the number of unique comp numbers across all derivatives

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
AdamG412: Not sure, this was here before I started. As per the comment on fillpatch we're still limited to contiguous components then I think it should still work

tamaraevst: This check in the assert only works for contiguous order. For example, you can end up querying components 1, 5, 8, giving you 3 components in total. With this, the assert will fail.

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
tamaraevst: Yes, tracking unique comps will definitely help. Though I do not understand why the same will not help with non-contiguous order. Can we not just call `FillPatch` on the unique components that we need to interpolate? Do you need non-contiguous order for the AH finder?

AdamG412: Ah yeah we can just loop over and do them one by one. It's turned out that I don't actually need it to be non-contiguous, but I do have to maintain two particle interpolators because of this so I'll change it to fix that

Tests/ParticleInterpolatorUnitTest/ParticleInterpolatorUnitTest.cpp
AdamG412: good shout, it shows up a segfault

tamaraevst: Oh no! Did you merge modifications from PR #211 ?

AdamG412: ah no not yet, hopefully that will sort it

AdamG412: Was actually just me setting up the interpolator with the wrong number of comps - I'll add a proper error message in for that though since I imagine it might be quite a common occurence

PR #194, opened 2026-07-22
Add TwoPunctures initial data integration

### LINKED ISSUE #18: Port TwoPunctures initial data

mirenradia:
This PR adds optional TwoPunctures initial data to the BinaryBH example and resolves #18. It should be reviewed with GRTLCollaboration/TwoPunctures#7 and merged after that (hence why this is draft). As before, this repo is kept separate for licensing incompatibility reasons.

## Build instructions

To build the BinaryBH example with TwoPunctures support, you need to do the following:

1. Clone the [TwoPunctures](https://github.com/GRTLCollaboration/TwoPunctures) repo next to GRTeclyn (like AMReX). You can also clone this somewhere else but then you'll need to set the `TWOPUNCTURES_HOME` makefile variable appropriately.
2. Make sure you have GSL installed/available in your environment (e.g. load a `gsl` module). If it's installed in a non-standard place or the module doesn't set `CPATH` and `LIBRARY_PATH` environment variables, you can set `GSL_HOME` to its base installation directory.
3. `cd /path/to/GRTeclyn/Examples/BinaryBH`
4. Set `USE_TWOPUNCTURES=TRUE` at build time (and `TWOPUNCTURES_HOME` if necessary) e.g.
   ```
   make -j 4 USE_TWOPUNCTURES=TRUE
   ```
5. Note that `TwoPunctures` will be appended to the filename stem/build config e.g. `BinaryBH3d.gnu.MPI.TwoPunctures.ex`.

## Parameters

Below is the list of parameters with the values as in `params_test.txt`. They are all scoped by `two_punctures`. They should be self explanatory and are mostly the same as in GRChombo. Commented out values have their default value assigned to them.

```
two_punctures.verbose = true

# two_punctures.calculate_target_masses = true
# two_punctures.target_mass_plus  = 0.5
# two_punctures.target_mass_minus = 0.5
# two_punctures.adm_tol           = 1.0e-10

two_punctures.mass_plus  = 0.47008673687400732
two_punctures.mass_minus = 0.47008673687400732

two_punctures.offset_plus  = +2.0
two_punctures.offset_minus = -2.0

two_punctures.momentum_plus  = 0.0 +0.1 0.0
two_punctures.momentum_minus = 0.0 -0.1 0.0

two_punctures.spin_plus      = 0.0 0.0 0.0
two_punctures.spin_minus     = 0.0 0.0 0.0

# two_punctures.use_spectral_interpolation = false
# two_punctures.initial_lapse = "psi^n"
# two_punctures.initial_lapse_psi_exponent = -2.0

# two_punctures.num_points_A   = 30
# two_punctures.num_points_B   = 30
# two_punctures.num_points_phi = 16

# two_punctures.solver_tol    = 1.0e-10
# two_punctures.solver_maxit  = 5
# two_punctures.epsilon       = 1.0e-6
# two_punctures.tiny          = 0.0
# two_punctures.extend_radius = 0.0

# two_punctures.swap_xz = false
# two_punctures.do_residuum_debug_output = false
# two_punctures.do_initial_debug_output = false
```

> [!NOTE]
> In the case spectral interpolation is used (`two_punctures.use_spectral_interpolation = true`), the initial data interpolation is very slow and this is further exacerbated by the larger number of cells per MPI process one typically has when running on GPUs. It may take several hours. The easiest workaround is to compute the initial data in a CPU-only run with more MPI processes, write a checkpoint file and then restart on GPUs from that.

### `TwoPuncturesInitialData` class

* This object now stores a static `TP::TwoPunctures` object rather than this being stored in `TPAMR` (a child of `BHAMR`). The `TPAMR` class has been removed.
* Parameter loading is now done in the `TwoPuncturesInitialData::read_parameters()` function rather than in `SimulationParameters`. Note that this function is still called in `SimulationParameters` as we need to use the TwoPunctures parameters to set the normal BH parameters as these are used by the puncture tracker and puncture tagger. This is a bit hacky but I leave this for #1 to clean up.
* I have left in the old parameter checking code even though it's not used in case this can be implemented with #1. It is preprocessed out with `#if 0`.

### `BinaryBHLevel::initData()`

* TwoPunctures solving is done here. Note that `TwoPunctures::solve()` is idempotent so the solve is only done the first time it's called (i.e. on the coarsest level).
* Since TwoPunctures does not support interpolation on the GPU, the interpolation of data onto the grid is done on the CPU into a `MultiFab` which has a host memory arena. The data is then copied asynchronously to the usual state `MultiFab`.

### Regression Tests GitHub Action

* This has been refactored so there is another job that builds and runs the BinaryBH example with TwoPunctures support. New reference data has also been added for this case.
* Note that I haven't updated the GitLab action to run this case on the CSD3 Nvidia A100s as this pipeline is more serialized and I thought it wouldn't be worth make it even longer.
* Currently, it uses the `enhancement/grteclyn_support` branch of the `TwoPunctures` repo but I will need to change this once GRTLCollaboration/TwoPunctures#7 is merged.

### `Make.twopunctures`

* I have copied the format of the Makefile used in AMReX for linking with external libraries (e.g. PETSc, HDF5). This has been incorporated by creating a new GRTeclyn `Make.defs` file which calls the usual AMReX one and then loads `Make.twopunctures`.


KAClough (dismissed):
This looks good to me and in addition it works in practice as we have demonstrated! I think you wanted @julianakwan to take a look before merging but for me it is good to go. 

@SamuelBrady will need to look at the params more closely when he does his refactoring, but this is good for now.

I note that the default for `two_punctures.use_spectral_interpolation` should probably be true since false gives ugly grid effects in the constraints (although to be fair it doesn't seem to affect the waveform significantly).

(I will add the detail from this PR to the GRTeclyn documentation too - it is on my to do list.)

mirenradia:
I mainly wanted @julianakwan to review the changes to the CI pipelines but she's welcome to review the rest of it as well.

KAClough:
Update: docs added here - https://grtlcollaboration.github.io/GRTeclyn/two_punctures/.

I have also double checked and this gives exactly the profiles that GRChombo's implementation does for the same params, modulo different grids due to tagging.

mirenradia:
@KAClough, you'll need to re-review because I have force-pushed changes which dismissed your review.

KAClough (dismissed):
I am, once again, very happy with this

Source/BlackHoles/TwoPuncturesInitialData.hpp
```diff
@@ -9,47 +9,365 @@
...
+            s_two_punctures.center_offset[offset_dir] + s_two_punctures.par_b;
+        bh2_params.center[offset_dir] +=
+            s_two_punctures.center_offset[offset_dir] - s_two_punctures.par_b;
+    }
+
+    AMREX_FORCE_INLINE void
+    operator()(int ix, int iy, int iz,
+               const amrex::Array4<amrex::Real> &a_state) const
+    {
+        AMREX_ASSERT(s_two_punctures_solved);
```
julianakwan, line 78: Do we want `AMREX_ALWAYS_ASSERT` in case of non-debug builds?

Source/BlackHoles/TwoPuncturesInitialData.hpp
```diff
@@ -9,47 +9,365 @@
...
+        const amrex::CellData<amrex::Real> &state_cell_data =
+            a_state.cellData(ix, iy, iz);
+
+        for (int i = 0; i < NUM_CCZ4_VARS; ++i)
+        {
+            // Set only the non-zero components explicitly below
+            state_cell_data[i] = 0.0;
+        }
+
+        // CCZ4Vars vars(state_cell_data);
```
julianakwan: Do we still need this comment?

Source/BlackHoles/TwoPuncturesInitialData.hpp
```diff
@@ -9,47 +9,365 @@
...
+        // default to Taylor expansion interpolation as it is much faster
+        bool use_spectral_interpolation = false;
+        pp.queryAdd("use_spectral_interpolation", use_spectral_interpolation);
+        s_two_punctures.grid_setup_method =
+            (use_spectral_interpolation) ? "evaluation" : "Taylor expansion";
+
+        std::string initial_lapse = "psi^n";
+        pp.queryAdd("initial_lapse", s_two_punctures.initial_lapse);
+
+        if (s_two_punctures.initial_lapse != "twopunctures-antisymmetric" &&
```
julianakwan: This could look tidier with `AMREX_ALWAYS_ASSERT_WITH_MESSAGE`

Source/BlackHoles/TwoPuncturesInitialData.hpp
```diff
@@ -9,47 +9,365 @@
...
+        }
+
+        // spectral grid parameters
+        s_two_punctures.npoints_A = 30;
+        pp.queryAdd("num_points_A", s_two_punctures.npoints_A);
+        s_two_punctures.npoints_B = 30;
+        pp.queryAdd("num_points_B", s_two_punctures.npoints_B);
+        s_two_punctures.npoints_phi = 16;
+        pp.queryAdd("num_points_phi", s_two_punctures.npoints_phi);
+        if (s_two_punctures.npoints_phi % 4 != 0)
```
julianakwan: Same as above

Source/BlackHoles/TwoPuncturesInitialData.hpp
```diff
@@ -9,47 +9,365 @@
...
+        {
+            s_two_punctures.par_P_plus[i]  = momentum_plus[i];
+            s_two_punctures.par_P_minus[i] = momentum_minus[i];
+            s_two_punctures.par_S_plus[i]  = spin_plus[i];
+            s_two_punctures.par_S_minus[i] = spin_minus[i];
+        }
+
+        amrex::Print() << "The corresponding momenta are"
+                       << "\nP_plus = ";
+        FOR (i)
```
julianakwan: Should these outputs be suppressed according to the level of verbosity? As there is a `s_two_punctures.verbose` option already.

Examples/BinaryBH/params_test.txt
```diff
@@ -17,6 +17,46 @@ amr.plot_int = 8
 plot_interval = 8
 amr.derive_plot_vars = ALL
 
+#################################################
+# TwoPunctures initial data parameters
```
julianakwan, line 21: Why are there commented out parameters? There should be a separate config file if you want to test a different Two Punctures setup.

Examples/BinaryBH/BinaryBHLevel.cpp
```diff
@@ -70,12 +70,42 @@ void BinaryBHLevel::initData()
...
+                               state_new.DistributionMap(), state_new.nComp(),
+                               state_new.nGrowVect(), mf_info);
+#else
+    amrex::MultiFab &host_state = state_new;
+#endif
+
+#ifdef AMREX_USE_OMP
+#pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
+#endif
+    for (amrex::MFIter mfi(state_new, amrex::TilingIfNotGPU()); mfi.isValid();
```
julianakwan, line 92: I'm not a fan of the old style `MFIter`s ... is this necessary because we want Two Punctures to run on CPU even the rest of the code is running on GPUs?

julianakwan: Because you can have nice looking `ParallelFor`s with OpenMP on CPUs but that `ParallelFor` will only get selected if `AMREX_USE_GPU=FALSE`.

julianakwan (changes requested):
Thanks for this PR! I know you asked me only to review the CI but I looked at all the changes anyway. 

I think it's not necessary to have two versions of the BinaryBH regression test with and without Two Punctures. I would prefer to have a single regression test for the BinaryBH example with all the features included and a unit tests for each feature so the failure mode can be identified. This would mean adding a new unit test for the Two Punctures part though.

Examples/BinaryBH/params_test.txt
mirenradia: The commented out parameters are there just to indicate the default values. They can be uncommented and changed if desired. I would like to leave them.

Examples/BinaryBH/BinaryBHLevel.cpp
mirenradia: Yes, exactly. I don't think there's a way to use the MF `ParallelFor`s on CPUs if you've built with GPU support.

Source/BlackHoles/TwoPuncturesInitialData.hpp
mirenradia: I think not since that would slow the code down and it should only be necessary if you're doing development (when you should probably build with `DEBUG=TRUE`). Bear in mind that this function is called in every cell on the grid.

mirenradia:
> I think it's not necessary to have two versions of the BinaryBH regression test with and without Two Punctures. I would prefer to have a single regression test for the BinaryBH example with all the features included and a unit tests for each feature so the failure mode can be identified. This would mean adding a new unit test for the Two Punctures part though.

I'm actually not sure what I would do for a unit test for TwoPunctures hence why I went for the regression test. I guess we could compute the constraints and check that they're small but this is a bit weaker than the regression test. Most of my ideas for a unit test are essentially the same as a regression test but maybe just comparing the data at `t=0`. @KAClough, do you have any ideas for a unit test for TwoPunctures?

Source/BlackHoles/TwoPuncturesInitialData.hpp
mirenradia: Removed

Source/BlackHoles/TwoPuncturesInitialData.hpp
mirenradia: Done

Source/BlackHoles/TwoPuncturesInitialData.hpp
mirenradia: I used `AMREX_ALWAYS_ASSERT` as I felt it's clear enough without a message.

Source/BlackHoles/TwoPuncturesInitialData.hpp
mirenradia: OK, I've made this change. I think most people will turn verbosity on in any case.

KAClough:
> > I think it's not necessary to have two versions of the BinaryBH regression test with and without Two Punctures. I would prefer to have a single regression test for the BinaryBH example with all the features included and a unit tests for each feature so the failure mode can be identified. This would mean adding a new unit test for the Two Punctures part though.
> 
> I'm actually not sure what I would do for a unit test for TwoPunctures hence why I went for the regression test. I guess we could compute the constraints and check that they're small but this is a bit weaker than the regression test. Most of my ideas for a unit test are essentially the same as a regression test but maybe just comparing the data at `t=0`. @KAClough, do you have any ideas for a unit test for TwoPunctures?

I am happy with a regression test for now, and maybe we can change this later to a unit test if we can think of one. But I also don't know what we would do - for the constraints we would need to check convergence not just that it is small if we really want to be sure which is a bit messy. Anything else would just be a kind of regression test anyway.

mirenradia:
Another thing a unit test wouldn't do is also test how it integrates with the BinaryBH example so my preference is to leave it as a regression test.

julianakwan:
> Another thing a unit test wouldn't do is also test how it integrates with the BinaryBH example so my preference is to leave it as a regression test.

I don't think the unit test should be instead of the current regression test. I think it would be nice to have something that separates out Two Punctures from the rest of the BinaryBH example so that if there's a bug with Two Punctures we can easily isolate it. 

I don't have a great idea for a test, I was thinking of doing something like what we do for the `BSSNMatterTest` and output the initial conditions and do a `h5diff` on the new ones vs the old ones. We can leave it for later if you want to merge this in today.

julianakwan (dismissed):
Thanks for the changes!

I think the commented out params would be cleaner in a different file like `params_two_punctures.txt` or maybe even leaving a commented line like `# suggested two punctures params`.  But I don't feel too strongly about it so I'll let you decide if you want to merge this right away.

mirenradia:
> I don't have a great idea for a test, I was thinking of doing something like what we do for the `BSSNMatterTest` and output the initial conditions and do a `h5diff` on the new ones vs the old ones. 

I guess this basically is just a regression test but just on the initial data.

> We can leave it for later if you want to merge this in today.

Yes that would be ideal. Can you open an issue for the unit test?

> I think the commented out params would be cleaner in a different file like params_two_punctures.txt or maybe even leaving a commented line like # suggested two punctures params.

I don't think we want more parameter files given #107! I'll add a comment saying commented out parameters show their default values.

julianakwan (approved):
LGTM

Issue #122, opened 2025-06-10
Files to remove later


KAClough:
AlwaysInline.hpp
We should get rid of this file and just use AMREX_FORCE_INLINE instead. 

MultiLevelTask.hpp
Port or remove. This just allows output on the first t=0 step.

KAClough:
I think this can be closed as what is left to do is now replaced by #203.

Issue #203, opened 2026-08-11
New way to output data at t=0 rather than using MutiLevelTask.hpp


AreefW:
We might want to remove `MutiLevelTask.hpp` rather than porting it. We only need it to output the data at t=0 in `specificPostTimeStep()`, we could possibly have something implemented in `GRAMRLevel`.

mirenradia:
Maybe overriding the virtual function `GRAMRLevel::specific_post_init()` as I think @KAClough suggested on Slack makes the most sense.

The gotcha is that unlike `GRAMRLevel::specificPostTimeStep` which, because of subcycling is called from finest to coarsest level, `specific_post_init()` is called from coarsest to finest level so one needs to be careful if things need to be done on multiple levels. I think this was previously just computing diagostic variables in GRChombo and hopefully using the multi-level `Amr::derive()` functions I added in AMReX-Codes/amrex#4780 can be used to do that properly instead.

PR #204, opened 2026-08-11
Cleanup unused files 

AreefW:
This PR will (partially) resolve the issue #122. 

The `Source/GRTeclynCore/AlwaysInline.hpp` and `Source/Grids/Cell.hpp` have been moved to `Tests/CCZ4RHSTest` and `Tests/Common`, respectively. Note that outputting data at t = 0 using `MultiLevelTask.hpp` has been moved to the new issue #203.


Tests/CCZ4RHSTest/Cell.hpp
mirenradia, line 1: I think this can be moved to the `CCZ4RHSTest` subdirectory and the include removed from `DerivativeTestsCompute.hpp`.

AreefW: Thanks for the comment `Cell.hpp` is now in `CCZ4RHSTest`, and the include were also removed from `HarmonicTest.hpp` and `SphericalHarmonicTest.cpp`.

Source/Tagging/FixedGridsTagger.hpp
```diff
@@ -22,14 +21,15 @@ class FixedGridsTagger
     std::array<double, AMREX_SPACEDIM> m_center;
 
   public:
-    // NOLINTBEGIN(bugprone-easily-swappable-parameters)
+    // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
     FixedGridsTagger(const double dx, const int a_level, const double a_L,
                      const std::array<double, AMREX_SPACEDIM> a_center)
         : m_dx(dx), m_L(a_L), m_level(a_level), m_center(a_center) {};
-    // NOLINTEND(bugprone-easily-swappable-parameters)
+    // NOLINTBEGIN(bugprone-easily-swappable-parameters)
```
mirenradia: Might as well just have one of these lines instead of two.
```suggestion
    // NOLINTBEGIN(bugprone-easily-swappable-parameters)
    FixedGridsTagger(const double dx, const int a_level, const double a_L,
                     const std::array<double, AMREX_SPACEDIM> a_center)
        : m_dx(dx), m_L(a_L), m_level(a_level), m_center(a_center) {};

```

mirenradia (changes requested):
Thanks for this PR. I have one very small suggestion.

Source/Tagging/FixedGridsTagger.hpp
AreefW: Thanks for the suggestion. Should we make the same change to other files as well?

For example, 
https://github.com/GRTLCollaboration/GRTeclyn/blob/293671e0ddab234e76525662d11a9af15a90c300/Source/Tagging/ChiTagger.hpp#L28

https://github.com/GRTLCollaboration/GRTeclyn/blob/293671e0ddab234e76525662d11a9af15a90c300/Source/CCZ4/PositiveChiAndLapse.hpp#L22

https://github.com/GRTLCollaboration/GRTeclyn/blob/293671e0ddab234e76525662d11a9af15a90c300/Source/Tagging/PunctureTagger.hpp#L32

and possibly adding them here too.

https://github.com/GRTLCollaboration/GRTeclyn/blob/293671e0ddab234e76525662d11a9af15a90c300/Source/Tagging/ExtractionTagger.hpp#L47

mirenradia: No, sorry. My suggested change was just because there are 3 `NOLINT` lines in this file where we can achieve the same thing with just 2 lines. Also, it's probably best to keep PRs focussed on a single task/issue and do unrelated changes in a different PR.

mirenradia (approved):
LGTM

Issue #171, opened 2026-03-26
Treatment of 8piG in the matter classes


KAClough:
Currently when we add matter to the RHS, we add the 8piG within the CCZ4RHSWithMatter class. In the new implementations of modified gravity we use this class to add in curvature terms that do not include the 8piG, so it might be better to have that added within the calculation of Tmunu. Otherwise we have to divide by 8piG and then multiply by it again, which is bad if you want to take the decoupling limit G=0.

KAClough:
This was addressed in #207 .

Issue #98, opened 2025-03-18
Redesign [Matter] Constraints class


mirenradia:
We should redesign these classes since we now use them in a different way to GRChombo because of the way derived variables work. Having the component numbers/intervals in the constructor doesn't really make sense any more.

We should allow the user to decide at compile time which components to store and whether or not to take the absolute values of the terms so that we can avoid runtime branching in the compute kernel.

KAClough:
Looking at the constraint code, I would favour always outputting just the two quantities Ham and M_i M^i, and the sum of the abs values for both. It is really rare that people want to focus on a specific direction and want all the mom constraints individually, and if they do want that they can hack it into their own version very easily. The current multiplicity of options is confusing and I think it is better to keep things simple in the public default.

KAClough:
Maybe we can also change the name `State_Type` to something less cryptic. Like `Temporary_State`? Or `Single_State`? Or just write above `base_index = 0` and then pass `base_index`? It's a very confusing way to provide the number 0.

mirenradia:
> Maybe we can also change the name `State_Type` to something less cryptic. Like `Temporary_State`? Or `Single_State`? Or just write above `base_index = 0` and then pass `base_index`? It's a very confusing way to provide the number 0.

I've opened #128 for this as it's more broad than just the constraint classes.

mirenradia:
> Looking at the constraint code, I would favour always outputting just the two quantities Ham and M_i M^i, and the sum of the abs values for both. It is really rare that people want to focus on a specific direction and want all the mom constraints individually, and if they do want that they can hack it into their own version very easily. The current multiplicity of options is confusing and I think it is better to keep things simple in the public default.

I agree that the current way to choose the different options is cumbersome but I think it's possible to design an interface that's easy to use with sensible defaults but _also_ allows an example developer to choose which they want to enable (at compile time).

Issue #20, opened 2023-02-13
Port ScalarField example


mirenradia:
This depends on #10.

KAClough:
This was done in #207 !

PR #207, opened 2026-08-12
Port ScalarField example with oscillaton

KAClough:
Here we update and port the scalar field example #20. We also fix the 8\piG  issue #171  by passing the full sources of the matter RHS rather than the emtensor components themselves. Note also #98. It now runs and gives a stable oscillaton, and is updated to the latest params changes.

Note that all matter classes are default constructable and I have removed the outdated load functions in GRParmParse as requested by @mirenradia in the params PR #215.

Docs could be more detailed but are complete.

@julianakwan I think it is better at this stage to make a separate PR for the CI changes you wanted to do, so this is good to review.


mirenradia:
I think we should wait until #1 is done before merging this and we should ensure it's consistent with the approach there.

Basically I don't want the extra constructor to `MatterCCZ4` and I want to remove arguments to class constructors and make things default-constructible where we can.

KAClough:
There is good news and bad news. The good news is that the example runs ok and generates nice profiles. The bad news is the oscillaton seems to be unstable and disperses. I suspect that somewhere the mass is being set to zero, but I will check in more detail tomorrow. 
lineouts

KAClough:
Fixed! 
Scalar-field lineouts

Note that this was caused by a bug in the way the emtensor was being computed when a potential term was present, so @julianakwan maybe we should update the tests to test this in future with a massive potential as the default.

julianakwan:
> Fixed! Scalar-field lineouts
> 
> Note that this was caused by a bug in the way the emtensor was being computed when a potential term was present, so @julianakwan maybe we should update the tests to test this in future with a massive potential as the default.

I don't know how that happened in the refractoring 😞 Maybe I copy and pasted the $-3 V(\phi)$ from somewhere? I hope it didn't take too long to fix :cry:

Yes, I agree the unit tests could be more robust so I will open an issue for this and the derivatives. Right now, only $\rho$ is being tested as well and I would like to extend the test to the other components.

Examples/BinaryBH/BinaryBHLevel.cpp
julianakwan, line 1: I think a lot of these changes are from branching off `enhancement/double_to_real` so I think we might need to rebase after PR #209 . It was a bit confusing at first but then I just skimmed over them.

julianakwan: By these files I mean the other example files.

Examples/ScalarField/Make.package
julianakwan, line 1: This should be removed with PR #224

Examples/ScalarField/OscillatonInitialData.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+// See the docs for further details on these initial conditions
+class OscillatonInitialData
+{
+  public:
+    AMREX_FORCE_INLINE
+    explicit OscillatonInitialData(amrex::Real a_dx);
+
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+    operator()(int ix, int iy, int iz,
```
julianakwan, line 30: Is this not throwing a linter error about easily swappable params? 😆

Examples/ScalarField/OscillatonInitialData.hpp
```diff
@@ -0,0 +1,70 @@
...
+  protected:
+    amrex::Real m_dx;
+    std::array<amrex::Real, AMREX_SPACEDIM> m_center{};
+
+    // These parameters come from fitting a known oscillaton solution
+    static constexpr amrex::Real geometry_scale       = 7.0;
+    static constexpr amrex::Real scalar_scale         = 10.0;
+    static constexpr amrex::Real scalar_central_value = -0.044159713033393284;
+
+    std::array<amrex::Real, 13> m_compactness_coefficients{
```
julianakwan: You may want to make these [AMReX literals](https://amrex-codes.github.io/amrex/docs_html/Basics.html#real)

Also is the magic number 13 necessary? I'm not sure, I think it may have come from an LLM fitting for these coefficients.

Examples/ScalarField/OscillatonInitialData.impl.hpp
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    // This will need to be set by the GammaCalculator, since the metric
+    // is not conformally flat, but we will zero it here for now.
+    FOR (i)
+    {
+        state_cell_data[c_Gamma1 + i] = 0.0;
+    }
+
+    // phi, K, A_ij, Theta, shift and B are all zero at this phase.
```
julianakwan: But should we explicitly set them to zero 🤔 ?

Examples/ScalarField/plot_lineouts.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python3
+
```
julianakwan, line 2: Either here or in the documentation, we should specify that the user needs to have `pathlib`, `matplotlib` and `numpy` installed, preferably in a virtual environment.

Examples/ScalarField/ScalarFieldLevel.cpp
```diff
@@ -3,115 +3,219 @@
...
-// For RHS update
-#include "MatterCCZ4RHS.hpp"
+#include "CCZ4RHSWithMatter.hpp"
+#include "ConstraintsWithMatter.hpp"
+#include "EMTensor.hpp"
+#include "FixedGridsTagger.hpp"
+#include "GRParmParse.hpp"
+#include "GammaCalculator.hpp"
+#include "LineExtraction.hpp"
+#include "MovingPunctureGaugeWithMatter.hpp"
```
julianakwan, line 15: Did you want to use the integrated puncture gauge for this example?

Examples/ScalarField/OscillatonInitialData.impl.hpp
julianakwan: Ah I see that it occurs later in `ScalarFieldLevel`

Examples/ScalarField/ScalarFieldLevel.cpp
```diff
@@ -3,115 +3,219 @@
...
+                           positive_chi_and_lapse(ix, iy, iz,
+                                                  soln_arrays[box_no]);
+                       });
+
+    GRParmParse evolution_pp("evolution");
+    int max_spatial_derivative_order{};
+    evolution_pp.get("spatial_derivative_order", max_spatial_derivative_order);
+    if (max_spatial_derivative_order != 4)
+    {
+        amrex::Abort("ScalarField currently supports fourth-order spatial "
```
julianakwan, line 151: Is there a reason for this? We do have sixth order derivatives now

Examples/ScalarField/StateVariables.hpp
```diff
@@ -7,25 +7,37 @@
...
-    ArrayTools::concatenate(CCZ4StateVariables::names, additional_names);
+    ArrayTools::concatenate(CCZ4StateVariables::names, scalar_field_names);
+
+static const std::array<BCParity, 2> scalar_field_parities = {BCParity::even,
+                                                              BCParity::even};
+
+static const std::array<BCParity, NUM_VARS> parities = ArrayTools::concatenate(
+    CCZ4StateVariables::parities, scalar_field_parities);
+
+static const std::array<amrex::Real, 2> scalar_field_asymptotic_values = {0.0,
```
julianakwan: This could be `NUM_VARS-NUM_CCZ4_VARS` to avoid the magic number

Source/BlackHoles/BinaryBHInitialData.hpp
julianakwan: I think we should avoid touching the files in the `BlackHoles` directory as well if the changes are only `double` to `amrex::Real`

Source/CCZ4/GammaCalculator.hpp
julianakwan, line 1: Should this be templated over the derivative class to enable sixth order derivatives?

Source/CCZ4/Weyl4.impl.hpp
```diff
@@ -55,8 +55,8 @@ Weyl4::operator()(int ix, int iy, int iz,
     weyl_scalar_t out = compute_Weyl4(ebfields, vars, h_UU, coords);
 
     // store the result
-    weyl_scalars(ix, iy, iz, m_out_comp)     = out.Real;
-    weyl_scalars(ix, iy, iz, m_out_comp + 1) = out.Im;
+    weyl_scalars(ix, iy, iz, m_dcomp)     = out.Real;
```
julianakwan, line 58: What was wrong with `m_out_comp`? I think we discussed this before as I was in favour of `dcomp` but was voted down 😂

Source/IO/GRParmParse.hpp
julianakwan, line 1: Why were those header files removed? Was it a linter thing?

Source/Matter/ScalarField.impl.hpp
```diff
@@ -64,6 +63,30 @@ AMREX_GPU_DEVICE emtensor_t ScalarField<potential_t, deriv_t>::compute_emtensor(
...
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE einstein_sources_t
+ScalarField<potential_t, deriv_t>::compute_einstein_sources(
+    int ix, int iy, int iz, const amrex::Array4<const amrex::Real> &state,
+    const deriv_t &a_deriv, const Tensor::Rank2 &h_UU) const
+{
+    const emtensor_t emtensor =
+        compute_emtensor(ix, iy, iz, state, a_deriv, h_UU);
+    const amrex::Real coupling = 8.0 * M_PI * m_G_Newton;
+
+    einstein_sources_t out;
```
julianakwan, line 76: I think it's confusing that now there are both `einstein_sources_t` and also `emtensor_t`. I know there is a comment in CCZ4Geometry.hpp but it would be much clearer if there was some explanation of the difference between the two in the docs as I wouldn't think to look for it here.

Source/ParticleInterpolator/LineExtractionParameters.hpp
```diff
@@ -0,0 +1,157 @@
...
+{
+    bool enabled{};
+    int num_points{};
+    std::array<amrex::ParticleReal, AMREX_SPACEDIM> start_coords{};
+    std::array<amrex::ParticleReal, AMREX_SPACEDIM> end_coords{};
+    std::string data_path;
+    std::string file_prefix;
+
+    explicit line_extraction_params_t(std::string a_param_scope)
+        : m_param_scope(std::move(a_param_scope))
```
julianakwan: I think `std::string` could be `std::string_view`. Why is the `std::move` necessary? Normally we just copy

Source/ParticleInterpolator/LineExtractionParameters.hpp
```diff
@@ -0,0 +1,157 @@
...
+        }
+        if (direction_norm_squared == 0.0)
+        {
+            extraction_pp.error("direction", "must be nonzero");
+        }
+
+        std::array<amrex::Real, AMREX_SPACEDIM> prob_extent{};
+        geometry_pp.get("prob_extent", prob_extent);
+        amrex::Real max_radius = 0.5 * prob_extent[0];
+        extraction_pp.queryAdd("max_radius", max_radius);
```
julianakwan: I find this setup a bit confusing (but maybe I am a bit slow...) why is there a radius and centre if we only want to extract the values in a line? It seems more natural to me to offer (x_start, y_start, z_start) and (x_end, y_end, z_end) as the parameters?

julianakwan (changes requested):
I only have a couple of minor comments. I ran the example so I can verify that it works but I haven't stress tested it yet. 

NB: I only went through the scalar field specific files carefully and kind of skimmed over the others as I think they were more to do with converting `double` -> `amrex::Real`.

Examples/BinaryBH/BinaryBHLevel.cpp
KAClough: Ah yes sorry, I wanted to start from that branch and I thought once it was merged the diff would work, but it seems not. I will fix it.

Examples/ScalarField/OscillatonInitialData.hpp
KAClough: Shhh! It hasn't noticed!

Examples/ScalarField/OscillatonInitialData.hpp
KAClough: Yes right, it is just the number needed for a good fit. I'll name it.

Examples/ScalarField/OscillatonInitialData.impl.hpp
KAClough: Yes, but let me add a comment to that effect here.

Examples/ScalarField/ScalarFieldLevel.cpp
KAClough: Yes, but I will leave that for @ShunhuiYao to implement after as he is the gauge guy.

Examples/ScalarField/ScalarFieldLevel.cpp
KAClough: Yes, again I wanted to leave it for @AreefW (the derivatives guy) so it is done consistently in both examples.

Source/CCZ4/GammaCalculator.hpp
KAClough: Yes, I will do this, even if we don't have sixth order ones in yet.

Source/CCZ4/Weyl4.impl.hpp
KAClough: I asked the machines to check the consistency of naming in our diagnostic files because we had changed things in some and not others. I personally still think `dcomp` is an abomination, but it was the majority case, and as I think you said it is what AMReX use, so I reluctantly reverted it...

Source/IO/GRParmParse.hpp
KAClough: They just weren't needed after the outdated load and default message code was removed I think.

Source/Matter/ScalarField.impl.hpp
KAClough: Ok, I will add something to the docs.

Source/ParticleInterpolator/LineExtractionParameters.hpp
KAClough: I think we just set this up as a toy example to test Tamara's code, but I agree these inputs make it more general. I can amend this.

Examples/BinaryBH/BinaryBHLevel.cpp
KAClough: Done

Examples/ScalarField/OscillatonInitialData.hpp
KAClough: Fixed

Examples/ScalarField/OscillatonInitialData.hpp
KAClough: Done

Examples/ScalarField/OscillatonInitialData.impl.hpp
KAClough: Done

Source/Matter/ScalarField.impl.hpp
KAClough: Done - I added a new matter section to the docs (and moved your diagnostic comments from the KG example into a new Diagnostics section too). I also updated the part on code structure now that this is mostly fixed. The machines also removed a load of trailing whitespace, and corrected my habitual mis-spelling of GRTeclyn, so it looks like I touched many files, but the code development ones were the only substantial ones.

Source/ParticleInterpolator/LineExtractionParameters.hpp
KAClough: Done, and move removed.

Source/ParticleInterpolator/LineExtractionParameters.hpp
KAClough: Done.

Examples/ScalarField/Make.package
KAClough: Done

Examples/ScalarField/plot_lineouts.py
KAClough: I added it here.

Examples/ScalarField/StateVariables.hpp
KAClough: Done

Source/CCZ4/GammaCalculator.hpp
KAClough: Done

docs/capabilities.md
```diff
@@ -24,7 +24,7 @@ Note that GRTeclyn is designed as a research tool, not a black box. Users are en
 
 ### Black hole spacetimes
 
-* Binary black hole (BBH) simulations are supported using Bowen–York initial data. 
+* Binary black hole (BBH) simulations are supported using Bowen–York initial data.
 * The **BinaryBH** example is fully functional for boosted cases. Currently the code uses an approximate solution but we plan to integrate the TwoPunctures code shortly, which will allow fully general initial momenta and spins.
```
julianakwan: Is TwoPunctures not already fully integrated?

Examples/ScalarField/plot_lineouts.py
julianakwan: Thanks! Does it require `pathlib` as well or is that part of `numpy`?

Source/CCZ4/GammaCalculator.hpp
```diff
@@ -6,57 +6,42 @@
...
+        const amrex::CellData<amrex::Real> &state_cell_data =
+            state.cellData(ix, iy, iz);
+        const amrex::CellData<const amrex::Real> &const_state_cell_data =
+            state_cell_data;
+        const CCZ4Vars vars(const_state_cell_data);
 
-        template <typename mapping_function_t>
-        AMREX_GPU_DEVICE void enum_mapping(mapping_function_t mapping_function)
+        const auto h_UU = CCZ4Geometry::compute_inverse_metric(vars);
+        const auto d1_h = m_deriv.d1_sym_tensor(ix, iy, iz, state, c_h11);
```
julianakwan: Aha! The `auto`s are convenient aren't they? 😆

Source/CCZ4/GammaCalculator.hpp
julianakwan: I think you might have to include the header as well.

julianakwan (changes requested):
Thanks for making those changes! I just have some minor comments this time then happy to merge.

docs/capabilities.md
KAClough: Ah yes, I wasn't aiming to edit this really, it just removed some whitespace, but then given that we are editing it maybe it is better to be up to date. I will fix it.

Examples/ScalarField/plot_lineouts.py
KAClough: According to the machines it is part of the python 3 standard library so if you have python you have this.

Source/CCZ4/GammaCalculator.hpp
KAClough: Ha! This was the lazy machines, I will fix it... at least for the d1 which is consistent with the RHS.

Examples/ScalarField/plot_lineouts.py
julianakwan: Great! Good to know

julianakwan (approved):
Thanks for the changes!

LGTM!

Issue #129, opened 2025-07-21
Replace hardcoded `double`s with `amrex::Real`s


mirenradia:
> Ask discussed with @julianakwan, we should go through and change doubles to amrex::Reals consistently (e.g. for y and z coords), so that we can easily test the switch to single precision if needed. This can be done is a separate PR to the removal of the data_t types.  

 _Originally posted by @KAClough in [#103](https://github.com/GRTLCollaboration/GRTeclyn/issues/103#issuecomment-3071001846)_

KAClough:
At least we should do this in the classes that are used in the Parallel4 loops on the device.

PR #208, opened 2026-08-12
Port SixthOrderDerivatives

### LINKED ISSUE #7: Port sixth order derivatives

AreefW:
This PR is for porting `SixthOrderDerivatives.hpp` to `GRTeclyn`. 

Once merged, we can then update the `BinaryBH` example with `SixthOrderDerivatives`

The PR will close #7.

### DIFF
--- a/Source/Grids/SixthOrderDerivatives.hpp
+++ b/Source/Grids/SixthOrderDerivatives.hpp
@@ -473,12 +473,12 @@ class SixthOrderDerivatives : protected DerivativeBase
         return advec_state;
     }
 
-    /*
-    // Eighth order dissipation: remember to change sign in front of factor in
-    // add_dissipation functions below if using this
+    // Eighth order dissipation
+    // Note the sign change before sigma_coeff, minus for 8th order and plus for
+    // 6th order
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE amrex::Real
-    dissipation_term_8th(const amrex::Real *in_ptr, const int stride,
-                         const int idx = 0) const
+    dissipation_term(const amrex::Real *in_ptr, const int stride,
+                     const int idx = 0) const
     {
         amrex::Real weight_vvfar = 3.906250e-3_rt;
         amrex::Real weight_vfar  = 3.125000e-2_rt;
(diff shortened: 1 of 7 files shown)


Source/Grids/SixthOrderDerivatives.hpp
```diff
@@ -9,496 +9,578 @@
...
-    static_assert(false, "SixthOrderDerivatives todo");
+  private:
+    amrex::Real m_dx;
+    amrex::Real m_one_over_dx;
+    amrex::Real m_one_over_dx2;
 
-  public:
-    const double m_dx;
+    [[nodiscard]] AMREX_GPU_DEVICE AMREX_FORCE_INLINE static const amrex::Real *
+    get_var_ptr(const int ivar, const amrex::Real *state_ptr_xyz,
```
julianakwan: Can you please make this a member function of a parent class or in a common header file? (First one is better practice by the way).

It is not great to have duplicated functions and I already defined these for `FourthOrderDerivatives`. 

Thanks!

Source/Grids/SixthOrderDerivatives.hpp
```diff
@@ -9,496 +9,578 @@
...
+                const amrex::GpuArray<int, AMREX_SPACEDIM + 1> strides) noexcept
+    {
+        return state_ptr_xyz + ivar * strides[3];
+    }
 
-  private:
-    const double m_one_over_dx;
-    const double m_one_over_dx2;
+    [[nodiscard]] AMREX_GPU_DEVICE AMREX_FORCE_INLINE
+        amrex::GpuArray<int, AMREX_SPACEDIM + 1> static get_strides(
```
julianakwan: (As above)

Source/Grids/SixthOrderDerivatives.hpp
```diff
@@ -9,496 +9,578 @@
 #include "DimensionDefinitions.hpp"
 #include "StateVariables.hpp"
 #include "Tensor.hpp"
+#include <AMReX_REAL.H>
 #include <array>
+#include "AMReX_Array.H"
```
julianakwan: This is already included in Tensor.hpp so not necessary to include again.

julianakwan (changes requested):
Thanks @AreefW for this PR! I have left a few comments for changes but I think it is a great start! 

Mainly, I think we might want to make a parent class for derivatives in general and then fourth and sixth order derivatives can inherit from to avoid duplication. Note that virtual functions can slow down the GPU performance a lot but as long as you avoid these (and I think you can) it should be okay.

mirenradia:
Would you also be able to update the Derivative Unit Tests accordingly?

AreefW:
Thank you both @julianakwan and @mirenradia, for the comments.

I'm going to create a parent class called `DerivativeBase` and will have both `get_var_ptr()` and `get_strides()` as member functions. Then `FourthOrderDerivative` and `SixthOrderDerivatives` can inherit from there.

I'll also update the Unit Tests after @julianakwan's comments are all resolved.

AreefW:
I've made a few changes to the branch.
- Switch to 8th-order dissipation suggested by @KAClough
- Switch `double` to `amrex::Real` following the #129 
- Update the `DerivativeUnitTest` for `SixthOrderDerivatives` with 8th-order dissipation
- Fix the `d1_tensor` bug following the #212

julianakwan:
Thank you for the changes! I think there are a few places where the derivatives can be tidied up by putting more functions in to the `DerivativesBase` class but let me think about it and get back to you. 

Also I think the unit test can be clean up as well as the fourth order and sixth order tests are essentially copy and pasted versions of each other.

mirenradia:
I have a suggested refactoring of the `*Derivatives` classes which you can find [here](https://github.com/GRTLCollaboration/GRTeclyn/compare/enhancement/sixth_order_derivatives...enhancement/derivatives_refactor_mr_suggest). I think this reduces code duplication in line with what Juliana was thinking.

I haven't touched the tests.

KAClough:
I would suggest that (now or in the future) we should change the tests so that they have a dependence on `ivar` - e.g. we could multiply the function by `ivar` in each case, which would then make sure we don't misorder the components (which is what we just corrected). It may also be better to use a function like `sin(xz)` rather than a simple polynomial in x and z, as in the case we are testing both fourth and sixth order stencils are exact.

AreefW:
Thank you @KAClough, for the suggestion. I think we might want to resolve those in a separate PR, as the problems were raised in #219.

Tests/DerivativeUnitTests/DerivativeUnitTests.cpp
```diff
@@ -87,174 +175,27 @@ void run_derivative_unit_tests()
             amrex::Gpu::streamSynchronize();
             AMREX_GPU_ERROR_CHECK();
 
-            constexpr amrex::Real test_threshold = 1e-10;
-
-            const auto &out_c_array = out_fab.const_array();
+            validate_derivatives(box, out_fab, dx, "fourth order", 1.0 / 64.0,
```
julianakwan: The magic numbers are not great practice. Could you please define a variable for the dissipation factor and power? Also add a comment to explain your choice of value. (Why is it 5?) Thanks!

Tests/DerivativeUnitTests/DerivativeUnitTests.cpp
```diff
@@ -87,174 +175,27 @@ void run_derivative_unit_tests()
...
-                                  "Failed advection down (fourth order) at ",
-                                  iv);
-
-                    CHECK_MESSAGE(cell_data[c_advec_up] ==
-                                      doctest::Approx(2. * z * (z - 1.) +
-                                                      3. * x * (2. * z - 1.))
-                                          .epsilon(test_threshold),
-                                  "Failed advection up (fourth order) at ", iv);
-                });
+            validate_derivatives(box, out_fab, dx, "sixth order", -1.0 / 256.0,
```
julianakwan: As above

julianakwan (changes requested):
Hi @AreefW - thanks for making those changes! I think the unit tests look good now, just some minor suggestions. We will of course refactor the derivatives at a later date.

AreefW:
@julianakwan I have made the changes as requested. Thank you for the suggestions, which keep the code written with good practices and make it clearer.

julianakwan (approved):
Thanks for the changes @AreefW !

I'm happy to merge this.

PR #211, opened 2026-08-13
Fix `ParticleInterpolator` bugs

### LINKED ISSUE #210: ParticleInterpolator bugs
Uncovered two important bugs in `ParticleInterpolator`:

* `amrex::AmrLevel::FillPatch` missing the last argument for `dcomp`. It defaults 0 so when `start_comp != 0`, the requested components are written into destination components starting at 0 rather than `start_comp`. The interpolation later reads components starting at `start_comp` in`interpolate_to_particle()`, so we can end up using the components that we did not fill in the ghosts for properly.

* `m_query` becomes stale after the execution of the query is complete and is not refreshed unless particles are being populated. This results in subsequent calls to `interp()` reusing a stale address.

tamaraevst:
This PR should address issue #210.

In particular, the bug change related to:
- Fixing args in `FillPatch()` of `ParticleInterpolator` (I was inattentive checking all the args of this function initially)
- query in `ParticleInterpolator` is now an argument for all functions that need it

In addition to the above I have added the following features:
- a couple of additional assert calls
- enforcing changing of `m_domain_ncell` when the number of levels changes

### DIFF
--- a/Source/ParticleInterpolator/ParticleInterpolator.hpp
+++ b/Source/ParticleInterpolator/ParticleInterpolator.hpp
@@ -148,8 +148,8 @@ class ParticleInterpolator
     // a_refresh_particles flag allows to "refresh" the particles, if e.g. the
     // user opts to change the query in the interpolation
     void interp(const InterpolationQueryParticle &query,
-                const std::string &name_derived = "", double time_derived = 0.0,
-                bool a_refresh_particles = false);
+                bool a_refresh_particles, const std::string &name_derived = "",
+                double time_derived = 0.0);
 
     void ensure_redistributed();
 
--- a/Source/ParticleInterpolator/ParticleInterpolator.hpp
+++ b/Source/ParticleInterpolator/ParticleInterpolator.hpp
@@ -43,9 +43,6 @@ class ParticleInterpolator
     amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> m_prob_lo{};
     amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> m_prob_hi{};
 
-    // number of cells per level
-    amrex::Vector<amrex::GpuArray<int, AMREX_SPACEDIM>> m_domain_ncell{};
-
     // reflective BC flags per side on the low and high sides
     amrex::GpuArray<bool, AMREX_SPACEDIM> m_lo_boundary_reflective{{false}};
     amrex::GpuArray<bool, AMREX_SPACEDIM> m_hi_boundary_reflective{{false}};
@@ -69,14 +66,15 @@ class ParticleInterpolator
     BoundaryConditions::params_t m_bc_params;
 
     // for getting the starting component of query
-    int get_start_comp(InterpolationQueryParticle &query);
-    int m_num_query_points{}; // for storing number of query points (later used
-                              // in the minimal check to verify that the query
-                              // has not changed).
+    int get_start_comp(const InterpolationQueryParticle &query);
+    std::size_t m_num_query_points{}; // for storing number of query points
+                                      // (later used in the minimal check to
+                                      // verify that the query has not changed).
     // it is possible that the user uses the same number of points, but the
-    // actuall query coords are different, but this would require a more complex
-    // infrastructre we also do not really carefully check for ordering of comps
-    // and assume the user knows what they are doing
+    // actual query coords are different, but this would require a more complex
+    // infrastructre -- to be considered in the future.
+    // we also do not really carefully check for ordering of comps and assume
+    // the user knows what they are doing -- this will be changed in Adam's PR.
 
     // mpi stuff
     MPIContextParticle m_mpi;
@@ -109,7 +107,7 @@ class ParticleInterpolator
 
     // A helper function that aggregates all the points together from senders
     // and receivers, collects the them into out arrays and applies parity
-    void aggregate_points(InterpolationQueryParticle &query);
+    void aggregate_points(const InterpolationQueryParticle &query);
 
     // A helper function to prepare send buffers, packs m_answer_idx and
     // m_answer_data
@@ -123,7 +121,7 @@ class ParticleInterpolator
     void exchange_answers();
 
     // Apply parities and store interpolated values in out arrays
-    void apply_parity_and_store_values(InterpolationQueryParticle &query);
+    void apply_parity_and_store_values(const InterpolationQueryParticle &query);
 
   public:
 
@@ -140,16 +138,18 @@ class ParticleInterpolator
                bool a_verbosity = false);
 
     // allocate particles at the query points
-    void populate_from_query(InterpolationQueryParticle &query);
+    void populate_from_query(const InterpolationQueryParticle &query);
 
     // A helper function that does interpolation from grid onto particles
     void interpolate_to_particle(int lev, amrex::MultiFab &mfab,
                                  const amrex::Geometry &geom, int start_comp);
 
     // final interpolation routine exposed to the users
-    void interp(InterpolationQueryParticle &query,
-                const std::string &name_derived = "",
-                double time_derived             = 0.0);
+    // a_refresh_particles flag allows to "refresh" the particles, if e.g. the
+    // user opts to change the query in the interpolation
+    void interp(const InterpolationQueryParticle &query,
+                const std::string &name_derived = "", double time_derived = 0.0,
+                bool a_refresh_particles = false);
 
     void ensure_redistributed();
 
--- a/Source/ParticleInterpolator/ParticleInterpolator.hpp
+++ b/Source/ParticleInterpolator/ParticleInterpolator.hpp
@@ -68,10 +68,15 @@ class ParticleInterpolator
     // copy of BC params
     BoundaryConditions::params_t m_bc_params;
 
-    // store the query here
-    InterpolationQueryParticle *m_query{};
     // for getting the starting component of query
-    int get_start_comp();
+    int get_start_comp(InterpolationQueryParticle &query);
+    int m_num_query_points{}; // for storing number of query points (later used
+                              // in the minimal check to verify that the query
+                              // has not changed).
+    // it is possible that the user uses the same number of points, but the
+    // actuall query coords are different, but this would require a more complex
+    // infrastructre we also do not really carefully check for ordering of comps
+    // and assume the user knows what they are doing
 
     // mpi stuff
     MPIContextParticle m_mpi;
@@ -104,7 +109,7 @@ class ParticleInterpolator
 
     // A helper function that aggregates all the points together from senders
     // and receivers, collects the them into out arrays and applies parity
-    void aggregate_points();
+    void aggregate_points(InterpolationQueryParticle &query);
 
     // A helper function to prepare send buffers, packs m_answer_idx and
     // m_answer_data
@@ -118,7 +123,7 @@ class ParticleInterpolator
     void exchange_answers();
 
     // Apply parities and store interpolated values in out arrays
-    void apply_parity_and_store_values();
+    void apply_parity_and_store_values(InterpolationQueryParticle &query);
 
   public:
 
@@ -135,11 +140,11 @@ class ParticleInterpolator
                bool a_verbosity = false);
 
     // allocate particles at the query points
-    void populate_from_query();
+    void populate_from_query(InterpolationQueryParticle &query);
 
     // A helper function that does interpolation from grid onto particles
     void interpolate_to_particle(int lev, amrex::MultiFab &mfab,
-                                 const amrex::Geometry &geom);
+                                 const amrex::Geometry &geom, int start_comp);
 
     // final interpolation routine exposed to the users
     void interp(InterpolationQueryParticle &query,
--- a/Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
+++ b/Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
@@ -296,8 +296,8 @@ void ParticleInterpolator<num_components>::interpolate_to_particle(
 // It uses/collates together all the methods defined in this class
 template <int num_components>
 void ParticleInterpolator<num_components>::interp(
-    const InterpolationQueryParticle &query, const std::string &name_derived,
-    double time_derived /*=0.0*/, bool a_refresh_particles /*=false*/)
+    const InterpolationQueryParticle &query, bool a_refresh_particles,
+    const std::string &name_derived, double time_derived /*=0.0*/)
 {
     AMREX_ASSERT(m_initialized);
 
--- a/Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
+++ b/Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
@@ -142,12 +142,10 @@ ParticleInterpolator<num_components>::reflect_particle(amrex::Real x,
 
 // allocate particles at the query points
 template <int num_components>
-void ParticleInterpolator<num_components>::populate_from_query()
+void ParticleInterpolator<num_components>::populate_from_query(
+    InterpolationQueryParticle &query)
 {
     AMREX_ASSERT(m_initialized);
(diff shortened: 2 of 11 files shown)


KAClough (commented):
This is fine for me but @mirenradia should approve.

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
```diff
@@ -270,6 +269,10 @@ void ParticleInterpolator<num_components>::interpolate_to_particle(
         const int num_particles = par_iter.numParticles();
         auto fab_array          = mfab[par_iter].const_array();
 
+        amrex::AllPrint() << "lev = " << lev << ", grid = " << par_iter.index()
```
KAClough: The machines say: This will print once per particle-containing grid on every extraction and could produce substantial output. I would remove it or guard it with m_verbosity before merging.

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
```diff
@@ -316,21 +321,20 @@ void ParticleInterpolator<num_components>::interp(
...
-        m_query = &query;
-
-        populate_from_query();
+        populate_from_query(query);
     }
 
-    AMREX_ASSERT(m_initialized);
+    AMREX_ALWAYS_ASSERT(query.numPoints() ==
+                        m_num_query_points); // check that the number of query
+                                             // points has not changed
```
mirenradia, line 336: You may not be surprised to hear that I'm not a fan of this not very robust check 😅.

Maybe it would just be easier to add an extra parameter to `interp()` which is `bool a_refresh_particles` (or something like that) where the user can be explicit whether they want to change the particles or not. If it's set to `true` then we'll need to remove the old particles and repopulate with the new ones.

Source/ParticleInterpolator/ParticleInterpolator.hpp
```diff
@@ -68,10 +68,15 @@ class ParticleInterpolator
     // copy of BC params
     BoundaryConditions::params_t m_bc_params;
 
-    // store the query here
-    InterpolationQueryParticle *m_query{};
     // for getting the starting component of query
-    int get_start_comp();
+    int get_start_comp(InterpolationQueryParticle &query);
+    int m_num_query_points{}; // for storing number of query points (later used
```
mirenradia: With my suggestion about the extra `interp()` argument, can we get rid of this? I think it's only used for the assertion.

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
mirenradia: I think it should be removed.

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
```diff
@@ -750,7 +762,20 @@ void ParticleInterpolator<num_components>::ensure_redistributed()
...
+
+        for (int lev = 0; lev < nlev; ++lev)
+        {
+            const amrex::Geometry &geom = m_gramr_ptr->getLevel(lev).Geom();
+
+            for (int d = 0; d < AMREX_SPACEDIM; ++d)
+            {
+                m_domain_ncell[lev][d] = geom.Domain().length(d);
+            }
+        }
```
mirenradia: Can we just remove `m_domain_ncell`? We can just get this from the `amrex::Geometry` object in `interpolate_to_particle` which I think is the only place where it is used.

mirenradia (changes requested):
Thanks for fixing these bugs. I have a couple of suggested changes below.

I'd also like you to make all of the `InterpolatorQueryParticle`s const references. I was able to do this by making the following changes.

<details>

<summary> Show diff </summary>

```diff
diff --git a/Source/ParticleInterpolator/InterpolationQueryParticle.hpp b/Source/ParticleInterpolator/InterpolationQueryParticle.hpp
index dbdd7bb7..5b5c6045 100644
--- a/Source/ParticleInterpolator/InterpolationQueryParticle.hpp
+++ b/Source/ParticleInterpolator/InterpolationQueryParticle.hpp
@@ -31,7 +31,7 @@ class InterpolationQueryParticle
 
     using comp_map_t = std::map<Derivative, std::vector<out_t>>;
     using iterator =
-        typename std::map<Derivative, std::vector<out_t>>::iterator;
+        typename std::map<Derivative, std::vector<out_t>>::const_iterator;
 
   private:
     template <int num_components> friend class ParticleInterpolator;
@@ -143,9 +143,9 @@ class InterpolationQueryParticle
 
     [[nodiscard]] size_t numPoints() const { return m_num_points; }
 
-    iterator compsBegin() { return m_comps.begin(); }
+    iterator compsBegin() const { return m_comps.cbegin(); }
 
-    iterator compsEnd() { return m_comps.end(); }
+    iterator compsEnd() const { return m_comps.cend(); }
 };
 
 #endif /* INTERPOLATIONQUERYPARTICLE_HPP_ */
diff --git a/Source/ParticleInterpolator/ParticleInterpolator.hpp b/Source/ParticleInterpolator/ParticleInterpolator.hpp
index 050d5a2f..10d1d6e6 100644
--- a/Source/ParticleInterpolator/ParticleInterpolator.hpp
+++ b/Source/ParticleInterpolator/ParticleInterpolator.hpp
@@ -69,7 +69,7 @@ class ParticleInterpolator
     BoundaryConditions::params_t m_bc_params;
 
     // for getting the starting component of query
-    int get_start_comp(InterpolationQueryParticle &query);
+    int get_start_comp(const InterpolationQueryParticle &query);
     int m_num_query_points{}; // for storing number of query points (later used
                               // in the minimal check to verify that the query
                               // has not changed).
@@ -109,7 +109,7 @@ class ParticleInterpolator
 
     // A helper function that aggregates all the points together from senders
     // and receivers, collects the them into out arrays and applies parity
-    void aggregate_points(InterpolationQueryParticle &query);
+    void aggregate_points(const InterpolationQueryParticle &query);
 
     // A helper function to prepare send buffers, packs m_answer_idx and
     // m_answer_data
@@ -123,7 +123,7 @@ class ParticleInterpolator
     void exchange_answers();
 
     // Apply parities and store interpolated values in out arrays
-    void apply_parity_and_store_values(InterpolationQueryParticle &query);
+    void apply_parity_and_store_values(const InterpolationQueryParticle &query);
 
   public:
 
@@ -140,14 +140,14 @@ class ParticleInterpolator
                bool a_verbosity = false);
 
     // allocate particles at the query points
-    void populate_from_query(InterpolationQueryParticle &query);
+    void populate_from_query(const InterpolationQueryParticle &query);
 
     // A helper function that does interpolation from grid onto particles
     void interpolate_to_particle(int lev, amrex::MultiFab &mfab,
                                  const amrex::Geometry &geom, int start_comp);
 
     // final interpolation routine exposed to the users
-    void interp(InterpolationQueryParticle &query,
+    void interp(const InterpolationQueryParticle &query,
                 const std::string &name_derived = "",
                 double time_derived             = 0.0);
 
diff --git a/Source/ParticleInterpolator/ParticleInterpolator.impl.hpp b/Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
index af9f63b2..50d68945 100644
--- a/Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
+++ b/Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
@@ -143,7 +143,7 @@ ParticleInterpolator<num_components>::reflect_particle(amrex::Real x,
 // allocate particles at the query points
 template <int num_components>
 void ParticleInterpolator<num_components>::populate_from_query(
-    InterpolationQueryParticle &query)
+    const InterpolationQueryParticle &query)
 {
     AMREX_ASSERT(m_initialized);
 
@@ -307,7 +307,7 @@ void ParticleInterpolator<num_components>::interpolate_to_particle(
 // It uses/collates together all the methods defined in this class
 template <int num_components>
 void ParticleInterpolator<num_components>::interp(
-    InterpolationQueryParticle &query, const std::string &name_derived,
+    const InterpolationQueryParticle &query, const std::string &name_derived,
     double time_derived /*=0.0*/)
 {
     AMREX_ASSERT(m_initialized);
@@ -401,7 +401,7 @@ void ParticleInterpolator<num_components>::interp(
 // prepares the final out arrays from interpolation
 template <int num_components>
 void ParticleInterpolator<num_components>::aggregate_points(
-    InterpolationQueryParticle &query)
+    const InterpolationQueryParticle &query)
 {
     AMREX_ASSERT(m_initialized);
 
@@ -607,7 +607,7 @@ void ParticleInterpolator<num_components>::exchange_answers()
 // Build values at particle positions and apply parities
 template <int num_components>
 void ParticleInterpolator<num_components>::apply_parity_and_store_values(
-    InterpolationQueryParticle &query)
+    const InterpolationQueryParticle &query)
 {
 
     int start_comp = get_start_comp(query);
@@ -653,7 +653,7 @@ void ParticleInterpolator<num_components>::apply_parity_and_store_values(
          ++deriv_it)
     {
         using comps_t = std::vector<typename InterpolationQueryParticle::out_t>;
-        comps_t &comps = deriv_it->second;
+        auto &comps   = deriv_it->second;
 
         const Derivative &dkey = deriv_it->first;
 
@@ -737,7 +737,7 @@ void ParticleInterpolator<num_components>::check_domain(
 // A getter function to get the starting component from the query
 template <int num_components>
 int ParticleInterpolator<num_components>::get_start_comp(
-    InterpolationQueryParticle &query)
+    const InterpolationQueryParticle &query)
 {
     auto it = query.compsBegin();
     AMREX_ASSERT(it != query.compsEnd());
```

</details>

mirenradia:
Also, the GitLab CI failed but it didn't push its status to GitHub for some reason :thinking:.

Edit: I think it's because `m_num_query_points` is currently `private` and CUDA doesn't like private member accesses in the enclosing parent function of extended device lambdas.

mirenradia:
I'm now confused why the GPU build CI workflow passed on GitHub :thinking:

tamaraevst:
> Also, the GitLab CI failed but it didn't push it's status to GitHub for some reason 🤔.
> 
> Edit: I think it's because `m_num_query_points` is currently `private` and CUDA doesn't like private member accesses in the enclosing parent function of extended device lambdas.

I cannot seem to see GitLab CI in the checks though... Have we removed it/can I see it elsewhere?

Source/ParticleInterpolator/ParticleInterpolator.hpp
tamaraevst: Yes, we use `m_num_query_points` to check that the number of points has not changed in the query. If we do not want that, we can certainly remove it. So, with `bool a_refresh_particles` the user must still be aware that they should be using the same query unless they change the flag. Just to check we are okay with that, as we will not have any internal checks in `ParticleInterpolator` checking that the query has not changed?

mirenradia:
> I cannot seem to see GitLab CI in the checks though... Have we removed it/can I see it elsewhere?

Yeah I said this in my comment above. It's not been removed and I'm not sure why it didn't appear.

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
tamaraevst: Yeah, this is a good idea

Source/ParticleInterpolator/ParticleInterpolator.hpp
mirenradia: OK, let's keep your existing assertion as well for now. Could you open an issue asking us to consider implementing a more robust check that the query has changed.

tamaraevst: Shall we still put in `bool a_refresh_particles` in `interp` though?

mirenradia: Yes, let's have both.

tamaraevst:
Ok I have implemented the above changes in. The only thing I am being slow about is why `m_num_query_points` is causing issues. I think we only use it on the host, so there should not be issues on a GPU side... (or maybe my coffee has not kicked in) I am going to wait how the checks complete; if anyone knows of the way to access the gitlab CI error message, let me know; I would interested in seeing why it is complaining.

mirenradia:
> Ok I have implemented the above changes in. The only thing I am being slow about is why `m_num_query_points` is causing issues. I think we only use it on the host, so there should not be issues on a GPU side... (or maybe my coffee has not kicked in) I am going to wait how the checks complete; if anyone knows of the way to access the gitlab CI error message, let me know; I would interested in seeing why it is complaining.

I think this is just a CUDA restriction for the enclosing parent function of a GPU extended device Lambda. I'm pretty sure you had the same problem during the development of `ParticleInterpolator`.

Unfortunately, the GitLab pipeline does not seem to be triggering on this branch anymore and I have no idea why. Do you have access to an Nvidia GPU you can test on?

tamaraevst:
> > Ok I have implemented the above changes in. The only thing I am being slow about is why `m_num_query_points` is causing issues. I think we only use it on the host, so there should not be issues on a GPU side... (or maybe my coffee has not kicked in) I am going to wait how the checks complete; if anyone knows of the way to access the gitlab CI error message, let me know; I would interested in seeing why it is complaining.
> 
> I think this is just a CUDA restriction for the enclosing parent function of a GPU extended device Lambda. I'm pretty sure you had the same problem during the development of `ParticleInterpolator`.
> 
> Unfortunately, the GitLab pipeline does not seem to be triggering on this branch anymore and I have no idea why. Do you have access to an Nvidia GPU you can test on?

But we do not use `m_num_query_points` in the GPU lambda? Yes, I can recompile on a GPU to see if I see any error.

mirenradia:
> But we do not use `m_num_query_points` in the GPU lambda? Yes, I can recompile on a GPU to see if I see any error.

Correct but I don't think that matters to CUDA's restriction.

mirenradia:
I manually forced the pipeline to run using the GitLab CLI (`glab`).

mirenradia:
Of course now it passes :sweat_smile:

tamaraevst:
> Of course now it passes 😅

Lol, that was fast! I am still compiling on my local GPU node...

tamaraevst:
I also made some changes to the `ParticleInterpolator` and `SphericalExtraction` tests (sorry to spam this PR with more changes). I basically added the state variables we use in these tests more consistently (i.e. including them into `Common/StateVariables.hpp`) and setting the necessary parities there. I also changed one of the polynomials to be odd in $z$, as I wanted to have an explicit test case for parity implementation when symmetric BCs are used. This was partly motived when I started reviewing Adam's PR and I think these changes make our tests much cleaner. 

The above described changes also meant making some small amendements in `BSSNMatter` test, where the number of state vars has now changed with the new additions.

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
```diff
@@ -304,33 +296,52 @@ void ParticleInterpolator<num_components>::interpolate_to_particle(
 // It uses/collates together all the methods defined in this class
 template <int num_components>
 void ParticleInterpolator<num_components>::interp(
-    InterpolationQueryParticle &query, const std::string &name_derived,
-    double time_derived /*=0.0*/)
+    const InterpolationQueryParticle &query, const std::string &name_derived,
+    double time_derived /*=0.0*/, bool a_refresh_particles /*=false*/)
```
mirenradia: Can we make `a_refresh_particles` the second argument to this function and either remove the default value so the user is forced to specify whether to refresh the particles or not or change the default value to `true` which should always behave correctly even if the query is changed (although the performance will be worse as particles will need to be populated each time)?

mirenradia (changes requested):
Thanks for making the changes. I have one more small suggestion.

Source/ParticleInterpolator/ParticleInterpolator.impl.hpp
tamaraevst: Sure, no problem. This has now been implemented (the user needs to specify the arg).

mirenradia (approved):
LGTM

Issue #198, opened 2026-08-10
Separate out gauge class from `CCZ4RHS` class


mirenradia:
I'm just moving this to a separate issue as #106 was about a different gauge issue.

> Having thought about this, I suggest that we split the gauge out of the RHS class completely, so that it can be fully general and is not restricted by the templating. Although it uses rhs.Gamma it is otherwise independent of the RHS calculation and can be very bespoke to the problem. @julianakwan could do this in the current PR, or we can leave it to a future one. I had a moment where I thought about splitting out the matter too, but I think that makes sense to retain as a template, since it always adds in the same way to the RHS terms, only the calculation of the stress energy tensor and rhs is general, and we can accommodate that fine by passing the state. 

 _Originally posted by @KAClough in [#106](https://github.com/GRTLCollaboration/GRTeclyn/issues/106#issuecomment-4543776145)_

KAClough:
Fixed in #223

Issue #219, opened 2026-08-17
Improve coverage of unit tests


julianakwan:
This issue is about improving the coverage of the unit tests to make sure that all* of the functions are being tested. Some of the current tests are also not discriminating enough. 

There are at least 3 cases where this is true:

- [ ] The `EMTensor` test has $V(\phi) = 0$ and only the energy density is being tested. 

- [ ] The `DerivativesUnitTest` uses the same function for all the test cases `d1_scalar`, `d1_tensor`, `d2_tensor` etc. and the same component (33) so it won't pickup problems with `ivar`. 
- [ ] `IntegratedMovingPunctureGauge` has no unit test
- [ ] `TwoPunctures` has no unit test (Issue #206 )

I know that 100% coverage is not realistic, but I think we can at least cover things at either the unit test level or regression test level.

KAClough:
It would be useful to also add a test for the IntegratedMovingPunctureGauge, and one that checks the variable eta value.

PR #223, opened 2026-08-19
Gauge: Refactor to be separate from RHS

ShunhuiYao:
This PR will close the issue #198 . We have implemented the following features:
1. The `CCZ4RHS` and `CCZ4RHSWithMatter`  no longer have gauge class as template parameter.  Now, the `MovingPunctureGauge` and `IntegratedMovingPunctureGauge` have their own `calculate_rhs` functions to update the lapse and shift.
2. Both gauge classes can optionally use spatially varying  $\eta$ following  [Eq. (11) of Schnetter (2010)](https://arxiv.org/abs/1003.0859).
3. Documentation for gauge classes is added.


Source/CCZ4/MovingPunctureGauge.hpp
```diff
@@ -146,6 +201,44 @@ class MovingPunctureGauge
...
     }
+
+    /// Calculate the gauge RHS using the fully accumulated Gamma RHS.
+    /** All vacuum and matter contributions to rhs[c_Gamma1 + i] must be added
+     * before this function is called so that the B-field RHS uses the complete
+     * time derivative of the conformal connection functions. The gauge fixer
+     * is applied to the resulting B-field RHS before this function returns.
+     */
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+    calculate_gauge_rhs(int ix, int iy, int iz,
```
KAClough: Can we make this just an operator(), as for example here: https://github.com/GRTLCollaboration/GRTeclyn/blob/2e70a061caba641027075821e2be4dbcdcd9836b/Source/CCZ4/PositiveChiAndLapse.hpp#L30.

I think since this is quite short anyway, we don't need three functions, and you could just move the other two `rhs_gauge` and `gauge_fixer` into this one, to avoid needing to pass things around.

Source/CCZ4/MovingPunctureGauge.hpp
```diff
@@ -111,14 +127,53 @@ class MovingPunctureGauge
...
+        const amrex::Real fix_radius_squared =
+            m_params.fix_radius * m_params.fix_radius;
+        const amrex::Real fix_factor =
+            fix_radius_squared / (radius * radius + fix_radius_squared);
+        const amrex::Real f_of_r =
+            m_params.fix_coeff * fix_factor + (1.0 - m_params.fix_coeff);
+
+        FOR (i)
+        {
+            rhs_cell_data[c_B1 + i] *= f_of_r;
```
KAClough: The goal of this factor is to reduce the eta coefficient where gradients become small. So it should modify the value of eta before it is used, not be applied to B - I think you got this from my hacky attempt to add it in after the gauge RHS has already been calculated, sorry!

The reference is https://arxiv.org/abs/1003.0859.

Source/CCZ4/MovingPunctureGauge.hpp
```diff
@@ -111,14 +127,53 @@ class MovingPunctureGauge
...
-    MovingPunctureGauge() { m_params.fill_params(); }
+    MovingPunctureGauge(double a_dx) : m_deriv(a_dx), m_dx(a_dx)
+    {
+        m_params.fill_params();
+        GRParmParse geometry_pp("geometry");
+        geometry_pp.get("center", m_params.center);
+    }
+
+    /// suppress the B-field RHS far away.
+    /** Multiplies each B-field RHS component by
```
KAClough: I would remove the long comment and just provide the reference: https://arxiv.org/abs/1003.0859.

Source/CCZ4/MovingPunctureGauge.hpp
```diff
@@ -98,6 +105,15 @@ class MovingPunctureGauge
                     "eta",
                     "usually O(1/M_ADM) so typically O(1) in code units");
             }
+
+            bool enable_fixer = false;
```
KAClough: I think we should name this `enable_eta_cutoff`, then we can have `eta_cutoff_coeff`, and the radius should be `eta_cutoff_radius`.

Source/CCZ4/MovingPunctureGauge.hpp
```diff
@@ -40,6 +44,9 @@ class MovingPunctureGauge
                                        //! shift condition on/off
         amrex::Real eta; //!< The eta in \f$\partial_t B^i = \partial_t \tilde
                          //!\Gamma - \eta B^i\f$
+        amrex::Real fix_coeff{0.0};    //!< Coefficient for B-field RHS damping
```
KAClough: An alternative to doing this switch (which I now think is simpler) would be to instead read in the domain length L in the params, and set the eta_cutoff_radius to be larger than L^3 by default. Then the default value is off, and it is just turned on by making the radius smaller. This will make the logic of applying it to eta simpler.

Tests/CCZ4RHSTest/GaugeFixer.hpp
```diff
@@ -0,0 +1,52 @@
+/* GRTeclyn
```
KAClough: I don't think we should add this file or modify the test, as long as we use a default value in which the cutoff radius for eta is larger than the domain size (or for which it is switched off), it will give the same result as the old code.

In the longer term, we may want to update the comparison code so it uses more recent code, and at that point we can include this cut off.

Tests/CCZ4RHSTest/CCZ4RHSTest.cpp
```diff
@@ -87,11 +95,40 @@ void run_ccz4_rhs_test()
 
         amrex::Real sigma = 0.3;
 
-        CCZ4RHS<MovingPunctureGauge, FourthOrderDerivatives> current_ccz4_rhs{
-            dx};
+        CCZ4RHS<FourthOrderDerivatives> current_ccz4_rhs{dx};
+        MovingPunctureGauge<FourthOrderDerivatives> moving_puncture_gauge(dx);
+        IntegratedMovingPunctureGauge<FourthOrderDerivatives>
```
KAClough: Let's leave this as a test of MovingPunctureGauge, and try to modify it as little as possible. I don't think we should add the integrated case here, we can add it to the issue that exists #219 to say we need to increase the test coverage for it. The addition below just makes this test confusing and less readable.

Tests/BSSNMatterTest/BSSNMatterTest.cpp
```diff
@@ -145,21 +149,16 @@ void run_bssn_matter_test()
...
-        // calculate the matter contribution
+        // Calculate the matter contribution before the gauge update so that
+        // the B-field RHS uses the complete Gamma RHS.
         amrex::ParallelFor(
             out_mf,
             [=] AMREX_GPU_DEVICE(int ibox, int ix, int iy, int iz)
             {
                 current_ccz4_rhs.add_emtensor_rhs(
                     ix, iy, iz, out_mf_array[ibox], in_c_array[ibox]);
+                moving_puncture_gauge.calculate_gauge_rhs(
```
KAClough: This seems to be before the matter RHS is added?

Tests/BSSNMatterTest/BSSNMatterTest.cpp
```diff
@@ -101,15 +102,18 @@ void run_bssn_matter_test()
         pp.add("gauge.shift_advec_coeff", 0.0);
         pp.add("gauge.eta", 1.0);
 
+        std::array<amrex::Real, AMREX_SPACEDIM> center{};
```
KAClough: Was this needed?

Tests/BSSNMatterTest/BSSNMatterTest.cpp
ShunhuiYao: Yes, this is because the matter RHS does not affect the rhs of Gamma and we have already got
 the full rhs of Gamma after the current_ccz4_rhs.add_emtensor_rhs(...) call.

KAClough: Ah, I see, you are right. Can you just amend the comment in line 152 above to say emtensor contribution, not "matter contribution" which is a bit ambiguous. You could even add "Note that it does not depend on the matter rhs".

Source/CCZ4/MovingPunctureGauge.hpp
ShunhuiYao: I think we'd better not do this. This is because operator() in the IntegratedMovingPunctureGauge is designed to initialize the B field and it has nothing to do with the computation of RHS. So we should not make operator() the RHS calculator in the MovingPunctureGauge so that the two classes are consistent. But I can remove the `rhs_gauge` and `gauge_fixer`  and leave only the `calculate_gauge_rhs`.

KAClough: Yes, I agree they should be consistent, but you could also rename the operator() function in the Integrated one as `set_initial_B_from_Gamma()`, and then use the operator as the RHS in all cases. I think would be more readable than making the default operator() function do something unique to one class.

Source/CCZ4/IntegratedMovingPunctureGauge.hpp
```diff
@@ -19,74 +19,76 @@
...
 
-    /// Vars needed internally in 'compute'
-    template <class data_t> struct Vars
+    /// Store the initial integrated Gamma-driver RHS in B.
+    /** This makes the non-advective part of the initial shift RHS vanish. The
+     * B field is subsequently frozen by calculate_gauge_rhs(), preserving the
+     * subtraction throughout the evolution.
+     */
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
+    operator()(int ix, int iy, int iz,
```
KAClough: I think we should not use the operator() for this function, but give it an informative name like `set_initial_B_to_Gamma()`. Then we could make the rhs calculation use the operator() (as it is the main function of the class), but I'm also ok to leave that as a named function too.

Source/CCZ4/IntegratedMovingPunctureGauge.hpp
```diff
@@ -19,74 +19,76 @@
...
-
-        current_cell.store_vars(B, GRInterval<c_B1, c_B3>());
     }
 
-    template <template <class> class vars_t, class d2_vars_t>
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
-    rhs_gauge(vars_t<amrex::Real> &rhs, const vars_t &vars<amrex::Real>,
-              const vars_t<Tensor::Rank1> &d1, const d2_vars_t &d2,
-              const vars_t<amrex::Real> &advec) const
+    calculate_gauge_rhs(int ix, int iy, int iz,
```
KAClough: Can we rename this `calculate_rhs` as now it lives in the gauge it will be accessed as `moving_puncture_gauge.calculate_rhs()`.

Source/CCZ4/MovingPunctureGauge.hpp
```diff
@@ -7,12 +7,16 @@
 #define MOVINGPUNCTUREGAUGE_HPP_
 
 #include "CCZ4Vars.hpp"
+#include "Coordinates.hpp"
 #include "DimensionDefinitions.hpp"
+#include "FourthOrderDerivatives.hpp"
```
KAClough, line 12: We also need to include `SixthOrderDerivatives.hpp`

Source/CCZ4/IntegratedMovingPunctureGauge.hpp
```diff
@@ -19,74 +19,76 @@
...
+class IntegratedMovingPunctureGauge : public MovingPunctureGauge<deriv_t>
 {
   public:
-    using params_t = MovingPunctureGauge::params_t;
+    using base_t   = MovingPunctureGauge<deriv_t>;
+    using params_t = typename base_t::params_t;
 
-  protected:
-    params_t m_params;
+    IntegratedMovingPunctureGauge(double a_dx) : base_t(a_dx) {}
```
KAClough: double -> amrex::Real

Source/CCZ4/MovingPunctureGauge.hpp
```diff
@@ -111,25 +129,88 @@ class MovingPunctureGauge
...
     };
 
   protected:
     params_t m_params{};
+    deriv_t m_deriv;
+    amrex::Real m_dx;
 
   public:
-    MovingPunctureGauge() { m_params.fill_params(); }
+    MovingPunctureGauge(double a_dx) : m_deriv(a_dx), m_dx(a_dx)
```
KAClough: double -> amrex::Real

KAClough (commented):
This looks great! Very good docs! A few minor amendments then it is good to go.

Source/CCZ4/IntegratedMovingPunctureGauge.hpp
ShunhuiYao: I will rename the interface as `gauge.calculate_rhs(...) ` for both gauge classes.  This is the minimal change for the moment.

KAClough (approved):
LGTM, if the tests pass we can merge!

PR #224, opened 2026-08-21
Remove Make.package

### LINKED ISSUE #180: Remove `Make.package` files or enforce consistency between them
We inherited these `Make.package` files from AMReX but we always manually include all of the `*.cpp` and `*.hpp` files in each directory in them anyway so we could just remove them and automatically just include all `*.cpp` and `*.hpp` files in all of the directories we include. I think this is what we do in GRChombo.

In AMReX, there's more of a point to it as they have some bits where files are conditionally included e.g. [here](https://github.com/AMReX-Codes/amrex/blob/b8117c772be841192859f501c6b869573647b631/Src/AmrCore/Make.package#L17-L19):
```make
ifeq ($(USE_PARTICLES), TRUE)
  CEXE_headers += AMReX_AmrParGDB.H AMReX_AmrParticles.H
endif
```
We don't ever do this and I don't think we will in the future. If we still want to keep `Make.package` files, we should enforce (in CI) that all source files are present somewhere in them and they are formatted consistently.

Thanks @SamuelBrady for the suggestion of getting rid of them.

AreefW:
This PR will close #180.

All `Make.package` files are removed. The `GNUmake`s are now grabbing all sources and headers using wildcards instead of the previous dependence on `Make.package`. 

Note that the `/Tools/GNUMake/packages/Make.twopunctures` still contains `Make.package`, as the file itself is in [another repository](https://github.com/GRTLCollaboration/TwoPunctures/blob/main/Source/Make.package) and the relevant documentation still needs to be updated.


.github/workflows/docs.yml
julianakwan, line 1: Can you please rebase this branch onto `develop` rather than making the changes here? It keeps the git history clean. Thanks!

Examples/BinaryBH/GNUmakefile
```diff
@@ -30,10 +29,20 @@ src_dirs := $(GRTECLYN_SOURCE)/BlackHoles \
...
-src_pack := $(foreach dir, $(src_dirs), $(dir)/Make.package)
-include $(src_pack)
-CEXE_sources += $(GRTECLYN_CEXE_sources)
-CEXE_headers += $(GRTECLYN_CEXE_headers)
+# Find all source and header files in the src_dirs
+CEXE_sources += $(notdir $(foreach dir, $(src_dirs), $(wildcard $(dir)/*.cpp)))
+CEXE_headers += $(notdir $(foreach dir, $(src_dirs), $(wildcard $(dir)/*.hpp)))
+CEXE_headers += $(notdir $(foreach dir, $(src_dirs), $(wildcard $(dir)/*.impl.hpp)))
+
+# Example specific source files
```
julianakwan: You can remove having to list all of these if you add the current directory to `src_dirs` above, i.e. 

```
src_dirs := blah blah
            $(GRTECLYN_SOURCE)/Tagging \
            .
```

julianakwan: You might want to add a comment though to point out the dot because it's quite easy to miss.

Examples/KleinGordon/GNUmakefile
julianakwan, line 1: Same for this file.

julianakwan (changes requested):
Thanks for this PR @AreefW !

Looks good to me - just a few minor suggestions.

AreefW:
Quick update: I just rebased `develop`after `refactor params` has been merged and resolved a small conflict.

Examples/BinaryBH/GNUmakefile
AreefW: Thanks for pointing this out. I have made the change as suggested.

SamuelBrady: I've suggested changing the `.` to `$(realpath .)` and removing `$(realpath .)` from `INCLUDE_LOCATIONS` since this now includes it through `src_dirs`

AreefW: @SamuelBrady, I just pushed the changes for your comment.

Tests/GNUmakefile
```diff
@@ -43,16 +42,19 @@ test_dirs := $(patsubst %/, %, $(test_dirs_match))
 # Finally add in the Common directory
 test_dirs += $(GRTECLYN_TESTS_HOME)/Common
 
-src_pack := $(foreach dir, $(src_dirs), $(dir)/Make.package)
-test_pack := $(foreach dir, $(test_dirs), $(dir)/Make.package)
-include $(src_pack)
-include $(test_pack)
+search_dirs := . $(src_dirs) $(test_dirs)
```
SamuelBrady: This `.` should also be `$(realpath .)` if it's referring to the Tests directory itself

SamuelBrady (changes requested):
Looks good to me, only left one other comment. The meaning of `src_dirs` is now a little ambiguous, but I think that's ok, and I don't think there are any un-ported GRChombo cpp files (which would be getting compiled)

Tests/GNUmakefile
AreefW: Thanks for this. I just made the change.

SamuelBrady (approved):
LGTM!

julianakwan (approved):
Thanks for making those changes @AreefW !

I have also manually checked that there are no unwanted files that are being included by the grep. 

LGTM!

Issue #214, opened 2026-08-14
Remove CTO


KAClough:
Katy to remove the CTO in the CCZ4RHS as it creates a lot of confusion for users in the main BinaryBHLevel.cpp file, which is a high price to pay for something most people will not frequently use. 

(Sorry to Miren and Juliana who like it). 

Instead: Remove the CTO and the templating and use formulation/covariantZ4 and (formulation - 1.0)/(covariantZ4 - 1.0) as appropriate to select the correct expressions. Miren and Juliana are happy with this approach so long as the performance reduction is < 10% (which they both think it will be). The main disadvantage is that it will make CCZ4RHS less readable but Katy is willing to accept this.

KAClough:
Goodbye and good riddance. #228

PR #228, opened 2026-08-25
CCZ4RHS: Remove CTO

KAClough:
This is a PR to close issue #214.

It removed the CTO and replaces it with multiplication by 1.0 or 0.0 for the different branch options.

It is currently based off the double->Real branch, but will need to be rebased depending on when it goes in. 

For now it can be used to test the effect on performance for the profiling - @julianakwan maybe you want to check this?

### DIFF
--- a/Source/CCZ4/CCZ4RHS.impl.hpp
+++ b/Source/CCZ4/CCZ4RHS.impl.hpp
@@ -68,7 +68,6 @@ CCZ4RHS<gauge_t, deriv_t>::compute_chi_and_h_ij(
 }
 
 template <class gauge_t, class deriv_t>
-template <int formulation, int use_covariant_Z4>
 AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
 CCZ4RHS<gauge_t, deriv_t>::compute_A_ij_and_Theta_and_Gamma(
     int ix, int iy, int iz, const amrex::Array4<amrex::Real> &rhs,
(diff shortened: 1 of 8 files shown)


Source/CCZ4/CCZ4RHS.impl.hpp
```diff
@@ -196,56 +193,48 @@ CCZ4RHS<gauge_t, deriv_t>::compute_A_ij_and_Theta_and_Gamma(
...
-                (ricci.scalar + vars.K() * (vars.K() - 2.0 * vars.Theta())) -
-            kappa1_times_lapse * (double)GR_SPACEDIM * (1.0 + m_params.kappa2) *
-                vars.Theta() -
-            tr_covd2lapse -
-            2.0 * vars.lapse() * (double)GR_SPACEDIM /
-                ((double)GR_SPACEDIM - 1.0) * m_cosmological_constant;
-    }
+    const amrex::Real non_covariant_z4 = 1.0 - m_params.covariant_z4_coeff;
+    const amrex::Real kappa1_times_lapse =
+        m_params.covariant_z4_coeff * m_params.kappa1 +
```
julianakwan, line 198: This could be okay, but I believe `covariant_z4_coeff` is a `bool` so there is always an implicit cast being performed here. 

You may want to change the type of `covariant_z4_coeff` to `amrex::Real` but have the param being read in be a `bool`.

julianakwan (changes requested):
This looks good to me! I have checked that the performance remains roughly the same to within < 1% on all three architectures. 

I just had one small comment. I will look at it again when it has been rebased onto `develop` as sometimes it is a bit confusing with all the other changes.

KAClough:
Ok great, I will wait until the scalar field one is in develop, so I can update both at once, and then ask you to have another look.

Source/CCZ4/CCZ4RHS.impl.hpp
KAClough: I think this is fixed in the rebase, but let me know if it is not what you had in mind!

KAClough:
Ok @julianakwan I've rebased off the current main branch, and updated the commit so the matter CTO works too. Let me know what you think!

Source/CCZ4/CCZ4RHS.impl.hpp
julianakwan: Thanks, I think this is okay now.

julianakwan:
The GitLab pipeline that failed is just a time out. But I will try to manually check it anyway.

julianakwan (approved):
I have manually checked the failing CI. 

LGTM!

PR #230, opened 2026-08-28
Make examples support `SixthOrderDerivatives`

AreefW:
This PR adds support for 6th-order spatial derivatives to both the `BinaryBH` (vacuum) and `ScalarField` (matter) examples. 

** Changes**
* Updated `GRAMRLevel` to initialise `m_evolution_spatial_derivative_order`
* **BinaryBH Example Updates:** Implemented checks for `m_evolution_spatial_derivative_order` to run the correct template (4th or 6th order) in `BinaryBHLevel.cpp` inside `specificEvalRHS`.
* **ScalarField Example Updates:** Implemented similar changes for `ScalarFieldLevel.cpp` as in *BinaryBH Example*.

> **Note:** We can also update the `KleinGordon` example as well if needed.


KAClough:
In GRChombo we only templated over derivative type for the evolution and not for diagnostics, and I think we want to keep that in GRTeclyn. The diagnostics don't need to be more than 4th order accurate (since the particle extraction isn't more accurate than this), we just find the extra accuracy of 6th order important for the RHS evolution. So sorry Areef, but can you revert the diagnostic files?

Examples/BinaryBH/BinaryBHLevel.cpp
```diff
@@ -36,9 +37,25 @@ void BinaryBHLevel::variableSetUp()
...
+    else if (max_spatial_derivative_order == 6)
+    {
+        Constraints<SixthOrderDerivatives>::set_up(state_index);
+        Weyl4<SixthOrderDerivatives>::set_up(state_index);
+    }
+    // NOLINTEND(bugprone-branch-clone)
+    else
+    {
+        amrex::Abort("spatial_derivative_order must be 4 or 6");
+    }
```
SamuelBrady: This is already checked in BaseParameterChecker

Source/GRTeclynCore/GRAMRLevel.cpp
```diff
@@ -82,6 +82,8 @@ GRAMRLevel::GRAMRLevel(amrex::Amr &papa, int lev, const amrex::Geometry &geom,
 {
     GRParmParse pp;
     pp.get("evolution.nan_check", nan_check);
+    pp.query("evolution.spatial_derivative_order",
+             m_max_spatial_derivative_order);
```
SamuelBrady: This should be get, not query, and the BaseParameterChecker should handle defaulting it

Examples/BinaryBH/BinaryBHLevel.cpp
```diff
@@ -217,20 +229,44 @@ void BinaryBHLevel::specificEvalRHS(amrex::MultiFab &a_soln,
...
-            ccz4rhs.compute(rhs, state);
-        });
-#endif
+        CCZ4RHS<MovingPunctureGauge, SixthOrderDerivatives> ccz4rhs(
+            Geom().CellSize(0));
+
+        // NB: These are split up to avoid having to pre-compute all the
+        //  first and second derivatives in memory on the GPU at once.
+
+        amrex::ParallelFor(
```
julianakwan, line 228: Can you please hoist the `ParallelFor` loops outside of the `if` statement? They should be exactly the same as long as the `CCZ4RHS` object has been initialized with the right order of derivative. I would like to avoid the duplication if necessary.

Examples/BinaryBH/BinaryBHLevel.cpp
```diff
@@ -301,19 +337,50 @@ void BinaryBHLevel::tag_cells(amrex::TagBoxArray &a_tag_box_array,
...
+                    puncture_tagger(ix, iy, iz, tag_arrays[box_no]);
+                }
+            });
+    }
+    else if (m_max_spatial_derivative_order == 6)
+    {
+        ChiTagger<SixthOrderDerivatives> chi_tagger(Geom().CellSize(0),
+                                                    a_regrid_threshold);
+
+        amrex::ParallelFor(
```
julianakwan: Same for this loop.

Examples/BinaryBH/BinaryBHLevel.cpp
```diff
@@ -257,9 +293,11 @@ void BinaryBHLevel::pre_tag_cells()
     amrex::MultiFab &state_new = get_new_data(state_index);
     const auto current_time    = get_state_data(state_index).curTime();
 
-    // Just fill 2 ghosts for chi to calculate second derivatives
-    const int nghost = 2;
+    // Fill ghosts for chi to calculate second derivatives
+    // 4th-order d2 requires 2 ghost cells, 6th-order d2 requires 3 ghost cells
+    const int nghost = (m_max_spatial_derivative_order == 6) ? 3 : 2;
```
SamuelBrady: num_ghosts is already a parameter (with various checks) and can be overriden by the user, are we intentionally not using it here?

Examples/ScalarField/ScalarFieldLevel.cpp
```diff
@@ -35,8 +31,34 @@ void ScalarFieldLevel::variableSetUp()
 {
     BL_PROFILE("ScalarFieldLevel::variableSetUp()");
     stateVariableSetUp();
-    ScalarFieldConstraints::set_up(state_index);
-    ScalarFieldEnergyDensity::set_up(state_index);
+
+    // Since this is a static method, we parse the parameter directly
+    GRParmParse pp;
+    int max_spatial_derivative_order = 4;
+    pp.query("evolution.spatial_derivative_order",
```
SamuelBrady: This should be get, not query, and shouldn't be defaulted to four here - it should be the same as in BinaryBHLevel.cpp

Examples/ScalarField/ScalarFieldLevel.cpp
```diff
@@ -35,8 +31,34 @@ void ScalarFieldLevel::variableSetUp()
...
+                              SixthOrderDerivatives>::set_up(state_index);
+        EMTensor<ScalarFieldWithPotential<SixthOrderDerivatives>,
+                 EMTensorOptions::justEnergyDensity,
+                 SixthOrderDerivatives>::set_up(state_index);
+    }
+    // NOLINTEND(bugprone-branch-clone)
+    else
+    {
+        amrex::Abort("spatial_derivative_order must be 4 or 6");
+    }
```
SamuelBrady: This is already checked in BaseParameterChecker, and shouldn't be re-checked here

Examples/BinaryBH/BinaryBHLevel.cpp
AreefW: I just pushed new changes to get rid of the template in diagnostics, but that shouldn't have affected this part. 

I am not quite sure what I can do with `ParallelFor` here. Do you have suggestions?

julianakwan: I mean something like this:
```
if (m_max_spatial_derivative_order == 4)
{
CCZ4RHS<..., FourthOrderDerivatives> ccz4rhs(...)
}
elseif (m_max_spatial_derivative_order == 6)
{
CCZ4RHS<..., SixthOrderDerivatives> ccz4rhs(...)
}
else
{
amrex::Abort(...)
}

// Then the ParallelFors: 
ParallelFor(a_rhs, [=] ....)
{
ccz4rhs.compute_A_ij_and_Theta_and_Gamma(...)
}
```
Because none of the `ParallelFor`s depends on the order of the derivatives once the `ccz4rhs` object has been initialized so they should be outside of the if statement.

mirenradia: How would you ensure the `ccz4rhs` object is in scope for the `ParallelFor` kernel?

julianakwan: Ah yes, good point. You might have to do what we did for the [Klein-Gordon model](https://github.com/GRTLCollaboration/GRTeclyn/blob/main/Examples/KleinGordon/KleinGordonLevel.cpp). 

There might be a clever C++20 way to do it though.

AreefW:
> In GRChombo we only templated over derivative type for the evolution and not for diagnostics, and I think we want to keep that in GRTeclyn. The diagnostics don't need to be more than 4th order accurate (since the particle extraction isn't more accurate than this), we just find the extra accuracy of 6th order important for the RHS evolution. So sorry Areef, but can you revert the diagnostic files?

My bad, sorry for the mess! 🙏 

I've reverted the changes in diagnostics and squashed the commits.

Examples/BinaryBH/BinaryBHLevel.cpp
```diff
@@ -279,7 +279,7 @@ void BinaryBHLevel::pre_tag_cells()
 
     // Fill ghosts for chi to calculate second derivatives
     // 4th-order d2 requires 2 ghost cells, 6th-order d2 requires 3 ghost cells
-    const int nghost = (m_max_spatial_derivative_order == 6) ? 3 : 2;
+    const int nghost = (m_evolution_spatial_derivative_order == 6) ? 3 : 2;
```
KAClough: I don't think you need this any more, we just need 2 for 4th order.

Examples/ScalarField/ScalarFieldLevel.hpp
```diff
@@ -7,6 +7,7 @@
 #define SCALARFIELDLEVEL_HPP_
 
 #include "DefaultLevelFactory.hpp"
+#include "FourthOrderDerivatives.hpp"
```
KAClough, line 10: Also need sixth order include

Examples/BinaryBH/BinaryBHLevel.cpp
KAClough: Ok, I think we agreed to leave this for a future clever idea!

KAClough (dismissed):
LGTM

julianakwan:
I just merged in PR #231 from @llibert94 so you'll have to rebase this branch again.

AreefW:
> I just merged in PR #231 from @llibert94 so you'll have to rebase this branch again.

Ah, sorry, I merged main into this branch on the web before I saw your comment.  😅

KAClough (approved):

PR #231, opened 2026-08-28
Add det(h)=1 condition

llibert94:
This PR adds the det(h)=1 condition, which we checked it is essential to reproduce the expected results for the BinaryBH example.  The main change is to include this condition in the TraceARemoval class and change the name accordingly.
This PR addresses issue #101, although it is still pending to check whether we can get the same result with a damping in the RHS


KAClough:
This looks good to me but it is (I guess unavoidably) breaking the regression test. I think we will need advice from @julianakwan and @mirenradia on how to address this.

mirenradia:
> This looks good to me but it is (I guess unavoidably) breaking the regression test. I think we will need advice from @julianakwan and @mirenradia on how to address this.

You'll need to just update the regression test comparison plotfile under `.github/workflows/data`.

KAClough:
@llibert94 now @SamuelBrady has fixed these tests, I will run the full production example for some time and check that the waveform is comparable to the main branch for the first 100 or so time steps. Just to confirm that we haven't made a major error in how we amend the evolution. The error in the test was due to low resolution violating the algebraic constraint preservation, but in a well resolved simulation the difference should be small (although as we found it does make a cumulative difference over time).

KAClough:
I tested this against the develop branch, to check the evolutions were consistent. They are:

image

Note that the wiggles should be fixed by Llibert's next update.

KAClough (dismissed):
LGTM

julianakwan (changes requested):
Thanks for this PR @llibert94 ! 

Sorry to be a pain, but could you please remove the trailing white spaces in `plt00008_compare/Header`? If you have the pre-commit configured, it should flag it and automatically do it for you. 

Otherwise, looks good to me!

julianakwan (approved):
Thank you @llibert94 !

LGTM

PR #232, opened 2026-08-28
Fix minor errors in verbosity and docs

KAClough:
This PR updates the docs for a small typo and related errors.

It also makes the verbosity of the scalar field on, and fixes an inconsistency of whether the speed output depends on verbosity (now it does, to match the time output).


KAClough:
Ok I made a mess I give up!

PR #233, opened 2026-08-28
Make tagging flexible

llibert94:
This PR allows tagging to be more flexible. In particular, it includes two additional parameters in PunctureTagger.hpp and ExtractionTagger.hpp (level_separation and fudge_factor), which are defaulted to the values that have been tested for the BinaryBH example.


Source/Tagging/ExtractionTagger.hpp
```diff
@@ -25,8 +26,22 @@ class ExtractionTagger
...
+    amrex::Real m_level_separation{1.5};
 
   public:
+    static void check_params()
+    {
+        GRParmParse tagging_pp("tagging");
+        amrex::Real level_separation{1.5};
+        tagging_pp.get("level_separation", level_separation);
+
+        if (level_separation < 1.2)
```
KAClough: I think we should also check it is less than 2, as then it becomes very large.

Source/Tagging/PunctureTagger.hpp
```diff
@@ -24,8 +25,28 @@ template <unsigned int num_punctures> class PunctureTagger
         AMREX_SPACEDIM * num_punctures;
     std::array<amrex::Real, num_punctures> m_puncture_masses;
     std::array<amrex::Real, num_puncture_coords> m_puncture_coords;
+    amrex::Real m_level_separation{1.5};
+    amrex::Real m_fudge_factor{2.0};
```
KAClough: As much as I like fudge factor, we should probably call it `m_finest_level_factor`.

Source/Tagging/PunctureTagger.hpp
```diff
@@ -24,8 +25,28 @@ template <unsigned int num_punctures> class PunctureTagger
...
   public:
+    static void check_params()
+    {
+        GRParmParse tagging_pp("tagging");
+        amrex::Real level_separation{1.5};
+        tagging_pp.get("level_separation", level_separation);
+        amrex::Real fudge_factor{2.0};
+        tagging_pp.get("fudge_factor", fudge_factor);
+
+        if (level_separation < 1.2)
```
KAClough, line 41: As above, and < 2, and make these warnings.

Source/Tagging/ExtractionTagger.hpp
```diff
@@ -25,8 +26,22 @@ class ExtractionTagger
...
+        {
+            tagging_pp.error("level_separation",
```
KAClough: Maybe let's make this a warning, not an error, as in principle people could set it to 1 or less and it is ok.

Source/Tagging/PunctureTagger.hpp
```diff
@@ -51,9 +77,10 @@ template <unsigned int num_punctures> class PunctureTagger
         // (just the top level would be ok, but doing two ensures
         // the top levels are well spaced)
 
-        // we want each level to be double the innermost one in size
+        // we want each level to be level_separation times the innermost one in
```
KAClough: This comment should probably say "times the finer level above"

Source/Tagging/ExtractionTagger.hpp
```diff
@@ -25,8 +26,22 @@ class ExtractionTagger
     const int *m_extraction_levels_ptr{nullptr};
     std::array<amrex::Real, AMREX_SPACEDIM> m_center;
     int m_level;
+    amrex::Real m_level_separation{1.5};
 
   public:
+    static void check_params()
```
KAClough, line 32: You need to call these check_params() functions in the SimulationParameters.hpp file of the Example to make sure they get checked at runtime.

KAClough: You also need to add them to the params files (commented out with the default values, and the appropriate tagging scope)

e.g.

`# tagging.level_separation = 1.5`

Also, you could make the scopes here more specific, e.g. have

`# extraction_tagging.level_separation = 1.5`
`# puncture_tagging.level_separation = 1.5`

That way both can be chosen independently.

Source/Tagging/ExtractionTagger.hpp
```diff
@@ -45,6 +60,8 @@ class ExtractionTagger
             // Avoids conditionals in the kernel by setting num to 0
             m_num_extraction_radii = 0;
         }
+        GRParmParse tagging_pp("tagging");
+        tagging_pp.get("level_separation", m_level_separation);
```
KAClough: You need to make all of these ones with defaults:

`tagging_pp.queryAdd("level_separation", level_separation);`

Instead of `get`, this should clear some of the failed tests.

There were also some changes to the tests needed. In the end it was just easier if I pushed them - it was the codex suggestion so you might want to amend it a bit, but you get the idea - see my commit.

KAClough:
I also have to point out that my instruction for this task said "And document it" 🤣

Source/Tagging/PunctureTagger.hpp
```diff
@@ -51,9 +84,10 @@ template <unsigned int num_punctures> class PunctureTagger
         // (just the top level would be ok, but doing two ensures
         // the top levels are well spaced)
 
-        // we want each level to be double the innermost one in size
+        // we want each level to be level_separation times the finer level
+        // above
         const int exponent       = std::min(m_max_level - m_level - 1, 1);
```
KAClough: I think in our updated version we didn't apply this min - this should never be called on m_level=max level and on max_level-1 we want it to do the exact region specified - i.e. exponent should be zero so factor = 1.

It just depends on what we mean by the convention for the fudge factor but for me this is clearer.

Source/Tagging/PunctureTagger.hpp
```diff
@@ -24,8 +25,35 @@ template <unsigned int num_punctures> class PunctureTagger
...
+        }
+        if (level_separation > 2.0)
+        {
+            puncture_tagging_pp.warning("level_separation",
+                                        "levels are too large");
+        }
+        if (finest_level_factor < 1.0)
+        {
+            puncture_tagging_pp.warning("finest_level_factor",
+                                        "finest level must be bigger");
```
KAClough: More informative: "finest level should be placed outside the BH horizon".

Also sorry to change my mind, I think this one should actually be an error as it will really make a mess. It will also prevent negative values being passed.

Source/Tagging/PunctureTagger.hpp
```diff
@@ -24,8 +25,35 @@ template <unsigned int num_punctures> class PunctureTagger
...
+        amrex::Real level_separation{1.5};
+        puncture_tagging_pp.queryAdd("level_separation", level_separation);
+        amrex::Real finest_level_factor{2.0};
+        puncture_tagging_pp.queryAdd("finest_level_factor",
+                                     finest_level_factor);
+
+        if (level_separation < 1.2)
+        {
+            puncture_tagging_pp.warning("level_separation",
+                                        "levels must be more spaced out");
```
KAClough: More informative - "Levels may be too close together, which results in boundary errors reflecting. Either increase this value (or set n_proper to be larger)."

Sorry also thinking about it there should be an actual error if the value is <=1 as it would make the levels smaller on each level, so can you add an additional check for this?

Source/Tagging/PunctureTagger.hpp
```diff
@@ -24,8 +25,35 @@ template <unsigned int num_punctures> class PunctureTagger
...
+        }
+        if (level_separation > 2.0)
+        {
+            puncture_tagging_pp.warning("level_separation",
+                                        "levels are too large");
```
KAClough: "Levels are more than doubling around punctures, which may result in too much refinement."

Source/Tagging/ExtractionTagger.hpp
```diff
@@ -25,8 +26,27 @@ class ExtractionTagger
...
   public:
+    static void check_params()
+    {
+        GRParmParse extraction_tagging_pp("extraction_tagging");
+        amrex::Real level_separation{1.5};
+        extraction_tagging_pp.queryAdd("level_separation", level_separation);
+
+        if (level_separation < 1.0)
+        {
+            extraction_tagging_pp.warning("level_separation",
```
KAClough: Amend wording as in puncture tagger.

Tests/PunctureTrackerTest/PunctureTrackerLevel.cpp
```diff
@@ -139,10 +139,16 @@ void PunctureTrackerLevel::check_puncture_tagging()
...
-    const amrex::Real exponent     = std::min(max_level - Level(), 1);
-    const amrex::Real factor       = fudge_factor * std::pow(2.0, exponent);
+    GRParmParse puncture_tagging_pp("puncture_tagging");
+    amrex::Real level_separation{};
+    puncture_tagging_pp.get("level_separation", level_separation);
+    amrex::Real finest_level_factor{};
+    puncture_tagging_pp.get("finest_level_factor", finest_level_factor);
+
+    const int max_level        = get_gramr_ptr()->maxLevel();
+    const amrex::Real exponent = std::min(max_level - Level(), 1);
```
KAClough: May need to fix the min here too.

Tests/PunctureTrackerTest/params_test.txt
```diff
@@ -27,6 +27,11 @@ amr.max_grid_size = 16
 # Grid box side lengths must be divisible by this value
 amr.blocking_factor = 8
 
+# Keep the refinement geometry tested here independent of the application
+# defaults used by BinaryBH.
+tagging.level_separation = 2.0
```
KAClough: Update scope to `puncture_tagging`

KAClough (approved):
LGTM

PR #234, opened 2026-08-29
Bugfix/docs and verbosity

KAClough:
This fixes a few minor errors in the docs and the consistency of the verbosity flag for timing outputs (which are generally useful, so I have enabled them in the scalar example).


llibert94 (approved):
This also looks good to me! It's useful to have power to remove these messages.

PR #235, opened 2026-08-29
Fix puncture output

### LINKED ISSUE #169: PunctureTracker sometimes produces lots of punctures.dat.old.XXXXXXX files
I'm not sure how to reproduce this reliably but I have sometimes noticed lots of `punctures.dat.old.*` files even when I have run a simulation once in a directory (i.e. no restarts).

I'm guessing there is a race condition somewhere related to #109. If so, this would affect any use of `SmallDataIO`.

KAClough:
This should fix #169 for the multiple outputs of puncture data, which is caused by them being initialised on multiple levels.


llibert94 (approved):
Looks good to me ! No comments from my side, nice and easy fix for that issue

PR #240, opened 2026-09-01
Add stoic quotes

KAClough:
This PR makes it so that the code writes an uplifting or thought provoking stoic message when compilation passes or fails, and on the successful completion or failure of a job. 

It should only use tools we already have, and is sourced by an easy to add to list of quotes in a text file, from which a header is produced during the build.

I need to add information to the docs, and some more quotes, then it should be good to go.

I note that there is currently no option to turn it off...


mirenradia (changes requested):
Thanks for this PR! I think it looks pretty good 🙂 

I have a few suggestions:
1. For both the build and run quotes, could you print them in a box so that they're more prominent?
2. Currently no quote is printed in the case the code aborts prematurely. Could you take over signal handling from AMReX, print the stoic quote first and then call whatever AMReX's signal handler calls.
3. Could you add a build option to turn this feature off?

KAClough:
I think this works now!

image

mirenradia (changes requested):
Thanks for making the changes. As discussed, I'm still not observing the stoic quote when I hit a false assertion or `amrex::Abort` is called. It would be great to handle this case as well if it's possible.

Could you also wrap the quote and the bounding box at 80 chars so it fits nicely in a reasonable width terminal emulator?

Finally, I think all C++ source files should be under either `Source`, `Tests` or `Examples` so maybe we can move the header files to either `Source/GRTeclynCore` or `Source/IO`. I think the former makes the most sense but I know you thought this code was too frivolous to include there.

KAClough:
Sadly, trying to implement this was a bit of a major endeavour, and I concluded that maybe it was a step too far for what is indeed a frivolous change :-)

I have reverted to just giving messages on compilation, which I think is enough fun for now, and we can upgrade it later if someone can find a simpler fix. 

I've added a box in the docs, and some more quotes. I think we could squash and merge this now @mirenradia.

mirenradia (approved):
Even though I'm sad you've removed it at runtime, this LGTM.

PR #242, opened 2026-09-02
Cast between ParticleReal and Real types and use rt literal for constants

SamuelBrady:
Fix conversions between amrex::ParticleReal and amrex::Real in case different precisions are used, and use rt for constants that should be amrex::Reals

### DIFF
--- a/Examples/BinaryBH/BinaryBHLevel.cpp
+++ b/Examples/BinaryBH/BinaryBHLevel.cpp
@@ -149,8 +149,12 @@ void BinaryBHLevel::initData()
 #endif
 
         get_puncture_tracker().set_puncture_coords(
-            {bh1_params.center[0], bh1_params.center[1], bh1_params.center[2],
-             bh2_params.center[0], bh2_params.center[1], bh2_params.center[2]});
+            {static_cast<amrex::ParticleReal>(bh1_params.center[0]),
+             static_cast<amrex::ParticleReal>(bh1_params.center[1]),
+             static_cast<amrex::ParticleReal>(bh1_params.center[2]),
+             static_cast<amrex::ParticleReal>(bh2_params.center[0]),
+             static_cast<amrex::ParticleReal>(bh2_params.center[1]),
+             static_cast<amrex::ParticleReal>(bh2_params.center[2])});
         // can't call start_from_initial_punctures() because we need the full
         // AMR grid first
     }
@@ -308,7 +312,7 @@ void BinaryBHLevel::tag_cells(amrex::TagBoxArray &a_tag_box_array,
 
     constexpr auto num_puncture_coords =
         static_cast<std::size_t>(AMREX_SPACEDIM * num_punctures);
-    std::array<amrex::Real, num_puncture_coords> puncture_coords{};
+    std::array<amrex::ParticleReal, num_puncture_coords> puncture_coords{};
     const bool puncture_tracking_enabled =
         get_bh_amr_ptr()->puncture_tracking_enabled();
 
--- a/Source/ParticleInterpolator/SphericalGeometry.hpp
+++ b/Source/ParticleInterpolator/SphericalGeometry.hpp
@@ -22,7 +22,7 @@
 class SphericalGeometry
 {
   private:
-    std::array<amrex::ParticleReal, AMREX_SPACEDIM> m_center;
+    std::array<amrex::ParticleReal, AMREX_SPACEDIM> m_center{};
 
   public:
     SphericalGeometry()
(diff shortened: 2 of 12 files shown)


Examples/BinaryBH/BinaryBHLevel.cpp
```diff
@@ -317,14 +317,20 @@ void BinaryBHLevel::tag_cells(amrex::TagBoxArray &a_tag_box_array,
         puncture_coords = get_puncture_tracker().get_puncture_coords();
     }
 
+    std::array<amrex::Real, num_puncture_coords> puncture_coords_real{};
+    for (unsigned int i = 0; i < num_puncture_coords; ++i)
+    {
+        puncture_coords_real[i] = static_cast<amrex::Real>(puncture_coords[i]);
```
julianakwan: I would prefer to change the puncture coords to be `amrex::ParticleReal` consistently rather than introduce a new array. I think this would be cleaner. 

The puncture tagger will have be changed as well [here](https://github.com/GRTLCollaboration/GRTeclyn/blob/84e2514512b6f092e04f95e166c81361fd5ec215/Source/Tagging/PunctureTagger.hpp#L27) and [here](https://github.com/GRTLCollaboration/GRTeclyn/blob/84e2514512b6f092e04f95e166c81361fd5ec215/Source/Tagging/PunctureTagger.hpp#L74) and [here](https://github.com/GRTLCollaboration/GRTeclyn/blob/84e2514512b6f092e04f95e166c81361fd5ec215/Source/Tagging/PunctureTagger.hpp#L108)

Source/ParticleInterpolator/SphericalGeometry.hpp
```diff
@@ -85,9 +85,12 @@ class SphericalGeometry
 
         const amrex::ParticleReal cylindrical_radius = a_radius * sin(a_theta);
         const std::array<amrex::ParticleReal, AMREX_SPACEDIM> displacement{
-            AMREX_D_DECL(cylindrical_radius * cos(a_phi),
-                         cylindrical_radius * sin(a_phi),
-                         a_radius * cos(a_theta))};
+            AMREX_D_DECL(
+                static_cast<amrex::ParticleReal>(cylindrical_radius *
```
julianakwan, line 89: Why does this static cast need to happen? `cylindrical_radius`  `a_radius`, `a_phi` and `a_theta` are already  `amrex::ParticleReal`s

Source/Tagging/ExtractionTagger.hpp
```diff
@@ -58,7 +58,11 @@ class ExtractionTagger
                      const spherical_extraction_params_t &a_params)
         : m_dx(dx), m_level(a_level)
     {
-        m_center = a_params.center;
+
+        FOR (i)
+        {
+            m_center[i] = static_cast<amrex::Real>(a_params.center[i]);
```
julianakwan: Is the extract center always on a particle? I thought you could name any co-ord in the simulation volume, not necessarily on a particle? 

If this is true, then the centre should be an `amrex::ParticleReal`.

julianakwan: I think the punctures are always particles so that is why I want to make the centre be a `amrex::ParticleReal` consistently.

julianakwan (changes requested):
Thanks for checking this @SamuelBrady !

I think there are more `amrex::Real` -> `amrex::ParticleReal` changes in the Coordinates.hpp file. The class could be templated over the type so you can use `amrex::ParticleReal` coordinates or `amrex::Real` coordinates without static casting.

Source/ParticleInterpolator/SphericalGeometry.hpp
SamuelBrady: If you don't have this and use single precision for particles it gets upset about the narrowing conversion (which I assume happens because cos(float) returns a double)

julianakwan: Are you sure it is because of the cos? According to the [docs](https://en.cppreference.com/cpp/numeric/math/cos) it should return a float. You could try `cosf` to be sure.

Source/Tagging/ExtractionTagger.hpp
SamuelBrady: I don't think the extraction tagger has to be around a puncture, so I've made the center a Real

Source/ParticleInterpolator/SphericalGeometry.hpp
SamuelBrady: Yes, the error you get without the cast goes away if you use cosf

Examples/BinaryBH/BinaryBHLevel.cpp
SamuelBrady: Agreed, and done

PR #243, opened 2026-09-02
Update building on GPUs docs to make the infrastructure flags clearer

hoyc1:
The purpose of this MR is to update the `Building on GPUs` docs to make the infrastructure flags clearer. I am happy to modify the text depending on review.

### DIFF
--- a/docs/building_gpus.md
+++ b/docs/building_gpus.md
@@ -22,7 +22,7 @@ If you are on a login node, this will be a CPU node, but you may be able to modu
 ```
 module load hipcc/6.3amd
 ```
-If you are ssh-ing directly into the GPU node the compiler may be installed already by default, or you module load it as above. (TIP: You can always try to compile and see if it complains about not having the compiler.). It is generally recommended to compile the code on the same GPU infrasture that you intend on running on to prevent downstream issues.
+If you are ssh-ing directly into the GPU node the compiler may be installed already by default, or you module load it as above. (TIP: You can always try to compile and see if it complains about not having the compiler.). For some computing clusters, it may be recommended to compile the code on the same GPU infrasture that you intend on running on to prevent downstream issues. You may need to check with your system admins if this is required or not.
 
 If you want to run with MPI over multiple GPUs, you will also need an MPI distribution, but note that this **needs to have been built with support for the GPUs you are using**, and often that won't be the case. You may need to ask the system admins for guidance. On Cosma8 they have an openmpi module with specific support, so you can just do:
 ```
--- a/docs/building_gpus.md
+++ b/docs/building_gpus.md
@@ -39,8 +39,7 @@ This is the easy bit! Update your `Make.local-pre` to activate the appropriate o
 * Intel GPUs: set `USE_SYCL=TRUE`
 * Nvidia GPUs: set `USE_CUDA=TRUE`
 
-Note that this usually sets the other options by default, but sometimes you may need to override them.
-Our [Example configs for specific HPC systems](example_configs.md) are a good place to look for inspiration.
+These options usually set a default compiler (`COMP`) which should work most of the time and there should rarely be a need to override this. Our [Example configs for specific HPC systems](example_configs.md) are a good place to look for inspiration.
 
 You will also probably need to give it a specific flag for the GPU architecture, which you can google for (or hopefully find in the system docs). For example, your `Make.local-pre` may look something like:
 ```
--- a/docs/building_gpus.md
+++ b/docs/building_gpus.md
@@ -12,9 +12,9 @@ Fortunately (thank you AMReX and historic US government funding :pray:), AMReX t
 
 For:
 
-* AMD GPUs: you need to be using the HIP compiler
-* Intel GPUs: you need to use SYCL compiler
-* Nvidia GPUs: you need to use the CUDA compiler
+* AMD GPUs: you need to use the HIP backend
+* Intel GPUs: you need to use SYCL backend
+* Nvidia GPUs: you need to use the CUDA backend
 
 You can think of them all as being like CUDA, but with a different name.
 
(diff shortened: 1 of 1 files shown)


docs/building_gpus.md
```diff
@@ -10,13 +10,19 @@ One of the annoying things about GPUs is that there are 3 types related to the 3
 
 Fortunately (thank you AMReX and historic US government funding :pray:), AMReX takes care of all the pain of implementing the code so it works on all three architectures. But you still have to think about this when you compile and run on a particular one. You may need to ask a lot of questions to the system admin, and you will probably end up feeling confused and stupid. That's ok. It will be worth it when you see the speed up, and once you are set up things should run smoothly.
 
-For AMD GPUs, you need to be using the HIP compiler, for Intel it is SYCL and for Nvidia it is the more well known CUDA. You can think of them all as being like CUDA, but with a different name.
+For:
+
+* AMD GPUs: you need to be using the HIP compiler
+* Intel GPUs: you need to use SYCL compiler
+* Nvidia GPUs: you need to use the CUDA compiler
```
mirenradia: I don't think the wording here is quite correct since there is no such thing as the "SYCL compiler". For targeting Intel GPUs with SYCL, AMReX uses the Intel oneAPI DPC++/C++ compiler. I think it'll be more accurate to say the "GPU backend" e.g.

```suggestion
* AMD GPUs: you need to use the HIP backend
* Intel GPUs: you need to use SYCL backend
* Nvidia GPUs: you need to use the CUDA backend
```

docs/building_gpus.md
```diff
@@ -10,13 +10,19 @@ One of the annoying things about GPUs is that there are 3 types related to the 3
...
+
+You can think of them all as being like CUDA, but with a different name.
 
 If you are on a login node, this will be a CPU node, but you may be able to module load the GPU compiler, e.g. on cosma8 you do
 ```
 module load hipcc/6.3amd
 ```
-If you are ssh-ing directly into the GPU node the compiler may be installed already by default, or you module load it as above. (TIP: You can always try to compile and see if it complains about not having the compiler.)
+If you are ssh-ing directly into the GPU node the compiler may be installed already by default, or you module load it as above. (TIP: You can always try to compile and see if it complains about not having the compiler.). It is generally recommended to compile the code on the same GPU infrasture that you intend on running on to prevent downstream issues.
```
mirenradia: I don't think we need to say this. It's usually fine to build the code on the login nodes which probably won't have GPUs.

docs/building_gpus.md
```diff
@@ -27,7 +33,13 @@ The good news about MPI is that if you really get stuck, it may be that you can
...
 ## You need to tell AMReX to compile with GPU offload support
 
-This is the easy bit! Update your `Make.local-pre` to activate the appropriate options, with `USE_CUDA=TRUE` for Nvidia, `USE_HIP=TRUE` for AMD or `USE_SYCL` for Intel GPUs. Note that this usually sets the other options by default, but sometimes you may need to override them.
+This is the easy bit! Update your `Make.local-pre` to activate the appropriate options for the infrastructure you are using. For:
+
+* AMD GPUs: set `USE_HIP=TRUE`
+* Intel GPUs: set `USE_SYCL=TRUE`
+* Nvidia GPUs: set `USE_CUDA=TRUE`
+
+Note that this usually sets the other options by default, but sometimes you may need to override them.
```
mirenradia: I don't think it's clear what is mean by "the other options". Did you mean e.g. `COMP`?

Maybe it would be best to say that explicitly:

```suggestion
These options usually set a default compiler (`COMP`) which should work most of the time and there should rarely be a need to override this.
```

mirenradia (changes requested):
Thanks @hoyc1 for the PR.

I have a few suggestions.

docs/building_gpus.md
hoyc1: I think this makes sense. I was keeping the original language in the current docs, see e.g. [here](https://grtlcollaboration.github.io/GRTeclyn/building_gpus/):

> For AMD GPUs, you need to be using the HIP compiler...

I like the suggestion though. I will implement these changes.

docs/building_gpus.md
hoyc1: I know this is recommended by the system admins for the cluster I was working on. I am happy to remove this though.

docs/building_gpus.md
mirenradia: Yeah I know. Let's blame @KAClough!

docs/building_gpus.md
hoyc1: I left this sentence unchanged, see Section `You need to tell AMReX to compile with GPU offload support` [here](https://grtlcollaboration.github.io/GRTeclyn/building_gpus/). I think the original sentence was implying that when specifying e.g. `USE_CUDA=TRUE`, `USE_HIP` is set to False but I could be wrong. I am happy to implement your suggestion.

mirenradia: Ah yes. I'm going to blame @KAClough again.

docs/building_gpus.md
mirenradia: I think it depends on the particular HPC system but I don't think it's universal advice.

hoyc1: This makes sense. I have changed the text to read:

> For some computing clusters, it may be recommended to compile the code on the same GPU infrastructure that you intend on running on to prevent downstream issues. You may need to check with your system admins if this is required or not.

Are you happy with this, or should I just remove entirely?

mirenradia: Yes, I'm happy with that.

docs/building_gpus.md
KAClough: Ha! Well I am always happy if someone else writes the docs instead!

mirenradia:
@hoyc1, can I re-review this?

hoyc1:
@mirenradia, yes please re-review. Apologies I forgot to tag you when I addressed your comments.

mirenradia (approved):
LGTM

PR #250, opened 2026-09-02
Ah finder area

rc634:
Added AHGeometry class. 
Find the proper area of the horizon 2-surface using the surface integral of the induced 2d metric.
To be setup inside the AHFinder object.
Use refresh() and area() once the horizon is found.

Small changes to AHFinder were needed to accomodate this.

### DIFF
--- a/Source/AHFinder/AHGeometry.cpp
+++ b/Source/AHFinder/AHGeometry.cpp
@@ -149,7 +149,7 @@ QAB AHGeometry::q_ab(int i, int j) const
     QAB qLL;
     qLL(0, 0) = TensorAlgebra::compute_dot_product(e_th, e_th, gamma);
     qLL(0, 1) = TensorAlgebra::compute_dot_product(e_th, e_ph, gamma);
-    qLL(1, 0) = q(0, 1);
+    qLL(1, 0) = qLL(0, 1);
     qLL(1, 1) = TensorAlgebra::compute_dot_product(e_ph, e_ph, gamma);
 
     return qLL;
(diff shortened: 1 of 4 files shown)


Source/AHFinder/AHGeometry.cpp
```diff
@@ -0,0 +1,184 @@
...
+    const amrex::Real d_phi   = 2.0 * M_PI / m_num_phi;
+    m_d_theta_d_phi           = d_theta * d_phi;
+
+    set_f_gradients();
+}
+
+
+// used for finite difference derivatives on the surface
+// assigns 'off the grid' points for derivatives at the edge
+std::array<int, 4> AHGeometry::neighbours(int i, int j) const
```
AdamG412, line 30: If we have this here then we can remove the duplicate function from AHFinder.impl.hpp

Source/AHFinder/AHGeometry.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+    east = i * m_num_phi + (j + 1) % m_num_phi;
+    west = i * m_num_phi + (j - 1 + m_num_phi) % m_num_phi;
+
+    return {north, south, east, west};
+}
+
+// f(theta, phi) is already a scalar field native to the ring grid
+// df/dtheta, df/dphi directly calcaulted
+void AHGeometry::set_f_gradients()
```
AdamG412, line 57: Is this what I've been calling h? might be good to keep it consistent

AdamG412 (approved):

PR #251, opened 2026-09-02
gitignore: Add parameters_and_version.txt

mirenradia:
Also remove *.h5 since we don't generate HDF5 files.


KAClough (approved):

mirenradia:
It was a mistake to remove `*.h5` as they're created by the tests. I will re-add it in another PR.

PR #253, opened 2026-09-02
CI: Launch tests directly in GitLab CI

mirenradia:
make run was rebuilding the executable before and the job was sometimes running out of time.


julianakwan (approved):
Thank you for making this change!

LGTM

PR #255, opened 2026-09-03
Make restart from checkpoint clearer in BinaryBH example

hoyc1:
The purpose of this PR is to make the restart from checkpoint clearer in the BinaryBH example with a specific comment. I initially missed this option when trying to restart from checkpoint previously.


KAClough (approved):
LGTM, thank you for this! We can merge when the tests pass.

Issue #246, opened 2026-09-02
Create AGENTS.md file for LLM users


julianakwan:
We should create a `AGENTS.md` file to be consumed by Claude/Codex/your favourite LLM. This to inform the LLM about the codebase (and also to prevent the LLMs from forming the wrong opinions about the codebase) as when it first encounters the code, it will look for an `AGENTS.md` file. 

Older versions of Claude will look for `CLAUDE.md` so we should create a symlink from this to `AGENTS.md`.

IanHawke:
[GRTeclyn_AGENTS.md](https://github.com/user-attachments/files/31743429/GRTeclyn_AGENTS.md)

The attached is a Claude-generated file, built within the context of constructing an external matter code that uses GRTeclyn to evolve the spacetime.

PR #215, opened 2026-08-14
Refactor parameters

SamuelBrady:
Update of parameter parsing to use AMReX style.
Remaining tasks:
- [x] Reinstate just_check_params
- [x] Reinstate checks for two punctures parameters
- [x] Reinstate write_boundary_conditions
- [x] Fix frozen unit tests

Closes #1 and closes #107

PR #258, opened 2026-09-03
Tools: Create guidance file for LLMs

julianakwan:
This will add `AGENTS.md` and `CLAUDE.md` files for LLMs to refer to when working on the repository. 

`CLAUDE.md` just refers directly to `AGENTS.md`. Claude seems to prefer this to a symbolic link. 

This will close issue #246

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,66 @@
+# GRTeclyn Agents Guide
+
+## Overview
+
+GRTeclyn is a numerical relativity code for solving Einstein's equations using finite volume and adaptive mesh refinement methods. It is built on top of the [AMReX](https://github.com/AMReX-Codes/amrex) library but we use AMRLevel functions not the AMR functions directly. AMReX supplies the supporting functions for AMR, load balancing, the data containers, parallelization and GPU offloading.
+
+## Purpose & Personas
+
+- **GRTeclyn developers** - You are an expert in GPU programming. You follow a predictable, documented workflow that respects the coding conventions defined in the documentation in `docs/`. 
+- **GRTeclyn users** - You will provide accurate build/test/docs guidance sourced from the repo.
+- **Downstream module developers** — You are responsible for downstream repositories e.g. `GRFolres`, `GRDzhadzha`, that are built *on top of* GRTeclyn and pinned directly against its class/method names (`GRAmrLevel`, `SimulationParametersBase`, `StateVariables` enums, the `matter_t` interface `CCZ4RHSWithMatter` expects). As with GRTeclyn developers, but also be mindful of breaking GRTeclyn changes. 
+
+
+## Documentation
+
+Refer to the [website](https://grtlcollaboration.github.io/GRTeclyn/) for the code documentation. The website should be considered to be an authoritative source. 
+
+## Code structure
+
+The code is structured as follows:
+
+- Source: This is where the main code is located. It is not specific to a particular physical scenario. 
+- Tests: These are the unit tests run via [doctest](https://github.com/doctest/doctest). Place unit tests for new features here. 
+- Examples: These are specific physics examples using the underlying code. BinaryBH runs a binary black hole merger. ScalarField runs an oscillaton. KleinGordon solves the Klein-Gordon equation without GR.
+- Tools: Contains old doxygen documentation, the default top-level Makefiles called by the examples and the copyright info. 
+- external: Contains the header for doctest. 
+- docs: Documentation lives here. 
+- .github: GitHub actions responsible for CI/CD. Also contains comparsion data for CI/CD.  
+- .gitlab: GitLab pipeline that runs the regression tests on the A100 partition on CSD3. 
+
+
+## Recommended workflow
+
+When setting up new users to the code or on a new system/HPC cluster: 
+
+- Install [pre-commit](https://pre-commit.com/) including clang-format and clang-tidy. Refer to `.pre-commit-config.yaml` for the correct version number of both.
+- Git clone AMReX 
+- Git clone GRTeclyn. Prefer the GRTeclyn directory to live next to the AMReX directory not as a sub-directory or vice versa.
+- Make a new branch from `main` for changes.
+- Refer to the coding conventions in the documentation.
+- Manually run the unit and regression tests after each commit. 
+
+## Rules for pull requests
+
+- Never push directly to `main`. This is protected by branch protection rules also. 
+- New feature branches should be named `enhancement/my-feature`.
+- Bugfixes should be named `bugfix/github-issue-number-or-name`. If no relevant issue number exists, create an issue.
+- Use a descriptive title with less than 50 characters.
+- Write a descriptive message that is formatted to have less than 70 characters per line.
+- Require that all the CI/CD tests have passed, except for draft pull requests. 
+- Before opening a pull request, git pull from the remote to ensure that the branch is up to date. Rebase if necessary. 
+- Small commits e.g. changes for the linter, should be squashed or fixed up.
+- Remind the user to update the documentation with the pull request. Always update the documentation if changes made by the pull request also update information in the documentation. Always update the documentation if changes made by the pull request are breaking changes.
+
+
+## Rules for opening a new issue
+
+- Check if the branch is up to date with `main`.
+- Check if there is an existing open issue that covers the same topic.
+- Write a detailed report on how to reproduce the issue, including the build environment and flags used, and quote the GitHub SHA.
+
+
+## GPU Lambda Safety
+
+Identical caveat to AMReX's own `AGENTS.md`
+
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,2 @@
+
+@ AGENTS.md


AGENTS.md
```diff
@@ -0,0 +1,66 @@
+# GRTeclyn Agents Guide
+
+## Overview
+
+GRTeclyn is a numerical relativity code for solving Einstein's equations using finite volume and adaptive mesh refinement methods. It is built on top of the [AMReX](https://github.com/AMReX-Codes/amrex) library but we use AMRLevel functions not the AMR functions directly. AMReX supplies the supporting functions for AMR, load balancing, the data containers, parallelization and GPU offloading.
```
mirenradia, line 5: What do you mean by "AMRLevel" functions? I think you mean to say we use `Amr` rather than `AmrCore`? Also, we should probably use the correct capitalisation i.e. "AmrLevel" if we do want to refer to it.

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Purpose & Personas
+
+- **GRTeclyn developers** - You are an expert in GPU programming. You follow a predictable, documented workflow that respects the coding conventions defined in the documentation in `docs/`. 
+- **GRTeclyn users** - You will provide accurate build/test/docs guidance sourced from the repo.
+- **Downstream module developers** — You are responsible for downstream repositories e.g. `GRFolres`, `GRDzhadzha`, that are built *on top of* GRTeclyn and pinned directly against its class/method names (`GRAmrLevel`, `SimulationParametersBase`, `StateVariables` enums, the `matter_t` interface `CCZ4RHSWithMatter` expects). As with GRTeclyn developers, but also be mindful of breaking GRTeclyn changes.
```
mirenradia, line 11: `SimulationParametersBase` no longer exists following #215.

AGENTS.md
```diff
@@ -0,0 +1,66 @@
+# GRTeclyn Agents Guide
+
+## Overview
+
+GRTeclyn is a numerical relativity code for solving Einstein's equations using finite volume and adaptive mesh refinement methods. It is built on top of the [AMReX](https://github.com/AMReX-Codes/amrex) library but we use AMRLevel functions not the AMR functions directly. AMReX supplies the supporting functions for AMR, load balancing, the data containers, parallelization and GPU offloading.
+
+## Purpose & Personas
+
+- **GRTeclyn developers** - You are an expert in GPU programming. You follow a predictable, documented workflow that respects the coding conventions defined in the documentation in `docs/`.
```
mirenradia, line 9: Maybe clarify that you are an expert in GPU programming with AMReX?

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+## Purpose & Personas
+
+- **GRTeclyn developers** - You are an expert in GPU programming. You follow a predictable, documented workflow that respects the coding conventions defined in the documentation in `docs/`. 
+- **GRTeclyn users** - You will provide accurate build/test/docs guidance sourced from the repo.
+- **Downstream module developers** — You are responsible for downstream repositories e.g. `GRFolres`, `GRDzhadzha`, that are built *on top of* GRTeclyn and pinned directly against its class/method names (`GRAmrLevel`, `SimulationParametersBase`, `StateVariables` enums, the `matter_t` interface `CCZ4RHSWithMatter` expects). As with GRTeclyn developers, but also be mindful of breaking GRTeclyn changes. 
+
+
+## Documentation
+
+Refer to the [website](https://grtlcollaboration.github.io/GRTeclyn/) for the code documentation. The website should be considered to be an authoritative source.
```
mirenradia, line 16: I think it would make sense to just refer to the `docs` directory as specified above. LLMs prefer markdown anyway rather than rendered HTML.

AGENTS.md
```diff
@@ -0,0 +1,66 @@
+# GRTeclyn Agents Guide
+
+## Overview
+
+GRTeclyn is a numerical relativity code for solving Einstein's equations using finite volume and adaptive mesh refinement methods. It is built on top of the [AMReX](https://github.com/AMReX-Codes/amrex) library but we use AMRLevel functions not the AMR functions directly. AMReX supplies the supporting functions for AMR, load balancing, the data containers, parallelization and GPU offloading.
+
+## Purpose & Personas
+
+- **GRTeclyn developers** - You are an expert in GPU programming. You follow a predictable, documented workflow that respects the coding conventions defined in the documentation in `docs/`. 
+- **GRTeclyn users** - You will provide accurate build/test/docs guidance sourced from the repo.
```
mirenradia, line 10: I'm a bit confused by this? Isn't this a developer?

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+
+Refer to the [website](https://grtlcollaboration.github.io/GRTeclyn/) for the code documentation. The website should be considered to be an authoritative source. 
+
+## Code structure
+
+The code is structured as follows:
+
+- Source: This is where the main code is located. It is not specific to a particular physical scenario. 
+- Tests: These are the unit tests run via [doctest](https://github.com/doctest/doctest). Place unit tests for new features here. 
+- Examples: These are specific physics examples using the underlying code. BinaryBH runs a binary black hole merger. ScalarField runs an oscillaton. KleinGordon solves the Klein-Gordon equation without GR.
```
mirenradia, line 24: It might be worth clarifying that the tests and each example are individual programs that can be built separately?

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Documentation
+
+Refer to the [website](https://grtlcollaboration.github.io/GRTeclyn/) for the code documentation. The website should be considered to be an authoritative source. 
+
+## Code structure
+
+The code is structured as follows:
+
+- Source: This is where the main code is located. It is not specific to a particular physical scenario.
```
mirenradia, line 22: It might be worth saying here that these files are in some sense a library (though not something that is built as a dynamic library or a static archive) and they need to be included in an example to make a full GRTeclyn code.

mirenradia: Also, could you describe each subdirectory of `Source`?

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+
+The code is structured as follows:
+
+- Source: This is where the main code is located. It is not specific to a particular physical scenario. 
+- Tests: These are the unit tests run via [doctest](https://github.com/doctest/doctest). Place unit tests for new features here. 
+- Examples: These are specific physics examples using the underlying code. BinaryBH runs a binary black hole merger. ScalarField runs an oscillaton. KleinGordon solves the Klein-Gordon equation without GR.
+- Tools: Contains old doxygen documentation, the default top-level Makefiles called by the examples and the copyright info. 
+- external: Contains the header for doctest. 
+- docs: Documentation lives here. 
+- .github: GitHub actions responsible for CI/CD. Also contains comparsion data for CI/CD.
```
mirenradia, line 28: Maybe it's worth saying what CI pipelines there are? Also "comparsion" -> "comparison".

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+- docs: Documentation lives here. 
+- .github: GitHub actions responsible for CI/CD. Also contains comparsion data for CI/CD.  
+- .gitlab: GitLab pipeline that runs the regression tests on the A100 partition on CSD3. 
+
+
+## Recommended workflow
+
+When setting up new users to the code or on a new system/HPC cluster: 
+
+- Install [pre-commit](https://pre-commit.com/) including clang-format and clang-tidy. Refer to `.pre-commit-config.yaml` for the correct version number of both.
```
mirenradia, line 36: I don't think `clang-tidy` is used by pre-commit. Also, setting up pre-commit necessarily installs the correct versions of the checks so I don't think there's any point in saying this.

Maybe it's worth saying to always format code with clang-format (using the version specified in pre-commit) in case the user does the committing rather than the LLM (if pre-commit is not set up).

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+- Git clone AMReX 
+- Git clone GRTeclyn. Prefer the GRTeclyn directory to live next to the AMReX directory not as a sub-directory or vice versa.
```
mirenradia, line 38: Maybe worth saying it's only necessary to do these steps if not done already?

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+- Make a new branch from `main` for changes.
```
mirenradia, line 39: Maybe add that it's worth fetching the latest changes from GitHub first to ensure changes are based on the latest commit on `main`?

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+- Tools: Contains old doxygen documentation, the default top-level Makefiles called by the examples and the copyright info. 
+- external: Contains the header for doctest. 
+- docs: Documentation lives here. 
+- .github: GitHub actions responsible for CI/CD. Also contains comparsion data for CI/CD.  
+- .gitlab: GitLab pipeline that runs the regression tests on the A100 partition on CSD3. 
+
+
+## Recommended workflow
+
+When setting up new users to the code or on a new system/HPC cluster:
```
mirenradia, line 34: Maybe it's worth saying that modules may need to be loaded before building on an HPC cluster? Ask the user if it's not clear what queue/partition they want to run/develop on so that appropriate modules can be selected.

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+## Recommended workflow
+
+When setting up new users to the code or on a new system/HPC cluster: 
+
+- Install [pre-commit](https://pre-commit.com/) including clang-format and clang-tidy. Refer to `.pre-commit-config.yaml` for the correct version number of both.
+- Git clone AMReX 
+- Git clone GRTeclyn. Prefer the GRTeclyn directory to live next to the AMReX directory not as a sub-directory or vice versa.
+- Make a new branch from `main` for changes.
+- Refer to the coding conventions in the documentation.
+- Manually run the unit and regression tests after each commit.
```
mirenradia, line 41: I think it would be worth specifying when to run each type of test. If your modifications only touch a specific example, then there's no point re-running the unit tests. Likewise, if you're only adding a new test, there's no point re-running the regression tests.

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+- Make a new branch from `main` for changes.
+- Refer to the coding conventions in the documentation.
+- Manually run the unit and regression tests after each commit. 
+
+## Rules for pull requests
+
+- Never push directly to `main`. This is protected by branch protection rules also. 
+- New feature branches should be named `enhancement/my-feature`.
+- Bugfixes should be named `bugfix/github-issue-number-or-name`. If no relevant issue number exists, create an issue.
+- Use a descriptive title with less than 50 characters.
```
mirenradia, line 48: I don't think we need to limit PR/issue titles to < 50 chars. I think this should only apply to commit messages.

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+- Git clone GRTeclyn. Prefer the GRTeclyn directory to live next to the AMReX directory not as a sub-directory or vice versa.
+- Make a new branch from `main` for changes.
+- Refer to the coding conventions in the documentation.
+- Manually run the unit and regression tests after each commit. 
+
+## Rules for pull requests
+
+- Never push directly to `main`. This is protected by branch protection rules also. 
+- New feature branches should be named `enhancement/my-feature`.
+- Bugfixes should be named `bugfix/github-issue-number-or-name`. If no relevant issue number exists, create an issue.
```
mirenradia, line 47: Currently, we do specify underscores instead of hyphens in the coding conventions. I don't mind either but we should be consistent.

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+- Use a descriptive title with less than 50 characters.
+- Write a descriptive message that is formatted to have less than 70 characters per line.
```
mirenradia, line 49: Again, this is just for commit messages.

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+- Require that all the CI/CD tests have passed, except for draft pull requests.
```
mirenradia, line 50: Do we want to say when to open a draft PR?

AGENTS.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Rules for pull requests
+
+- Never push directly to `main`. This is protected by branch protection rules also. 
+- New feature branches should be named `enhancement/my-feature`.
+- Bugfixes should be named `bugfix/github-issue-number-or-name`. If no relevant issue number exists, create an issue.
+- Use a descriptive title with less than 50 characters.
+- Write a descriptive message that is formatted to have less than 70 characters per line.
+- Require that all the CI/CD tests have passed, except for draft pull requests. 
+- Before opening a pull request, git pull from the remote to ensure that the branch is up to date. Rebase if necessary.
```
mirenradia, line 51: I'm a bit confused by this. Do you mean to make sure you've fetched `main` to make sure it's up to date and then rebase onto that before opening the PR?

mirenradia (changes requested):
Thanks for adding this.

I have individual comments on lines below.

Maybe it would be worth adding some abbreviated build instructions for the tests and examples too to save the LLM from having to read all of the documentation on building?

AGENTS.md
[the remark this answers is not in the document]
mirenradia: Ah yes, I noticed this "finite volume" mistake too but forgot to comment on it.

AGENTS.md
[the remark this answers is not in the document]
mirenradia: I'm going to resolve this as I've already commented on it.

AGENTS.md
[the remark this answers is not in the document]
mirenradia: I've also commented about this above but Copilot has added some stuff about clang-tidy so will leave this open.

AGENTS.md
julianakwan: I just deleted this part because we use functions from both `Amr` and `AmrCore` directories. I just didn't want the LLMs to change the way that we interact with AMReX.

AGENTS.md
julianakwan: Removed

AGENTS.md
julianakwan: Done

AGENTS.md
julianakwan: Often people will run multiple agents and each agent will have a specific role.  The users section is for an agent running but not developing the code.

AGENTS.md
julianakwan: I think the LLMs are capable or reading the .yml files on their own.

Issue #256, opened 2026-09-03
Compiling with PRECISION=FLOAT


YL-codehub:
For GPU having no fp64 hardware support or simply for lighter runs, compiling with PRECISION=FLOAT fails at runtime.

Incoming fix:
- making sure all literals are fixed correctly
- requiring additional flag USE_SINGLE_PRECISION_PARTICLES=TRUE
- document such case.

mirenradia:
Isn't @SamuelBrady looking at this in #242?

PR #259, opened 2026-09-03
Bugfix: Allow fp32 only run

YL-codehub:
This PR will close issue #256  
- Missing _rt added
- some documentation on the topic.

### DIFF
--- a/docs/building_gpus.md
+++ b/docs/building_gpus.md
@@ -79,3 +79,33 @@ amrex.use_gpu_aware_mpi = 1
 !!! note "Stoic build quotes"
 
     After each build, GRTeclyn prints a random Stoic quote for either success or failure. Add quotes to `Tools/StoicQuotes/StoicQuotes.txt`, or set `USE_STOIC_QUOTES=FALSE` on the command line or in `Make.local-pre` to disable them.
+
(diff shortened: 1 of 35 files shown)


julianakwan:
Thanks @YL-codehub ! 

There might be some overlap with PR #242 so I will wait until that is merged to review.

docs/building_gpus.md
```diff
@@ -79,3 +79,33 @@ amrex.use_gpu_aware_mpi = 1
 !!! note "Stoic build quotes"
 
     After each build, GRTeclyn prints a random Stoic quote for either success or failure. Add quotes to `Tools/StoicQuotes/StoicQuotes.txt`, or set `USE_STOIC_QUOTES=FALSE` on the command line or in `Make.local-pre` to disable them.
+
+## You need to match the precision
```
KAClough, line 83: I don't think this documentation belongs here as most users will not need this - not having double precision is still (thankfully) fairly unusual. I would create a `precision.md` file for a separate `Precision` section that lives under "running the code" (added in the `mkdocs.yml` file).  Then we can add information about the precision requirements and observations about how it affects performance and accuracy.

PR #264, opened 2026-09-11
Bugfix/bbhexample

ShaunSwain12:
In ```Examples/BinaryBH/params_production.txt```, only the TwoPunctures initial data parameters are specified. However, ```BinaryBHLevel.cpp``` explicitly attempts to read in ```bh1.mass``` and ```bh2.mass``` for the puncture tracker.

The fix is to add a branch that uses the TwoPunctures target or bare masses (whichever is specified by the user) if ```USE_TWOPUNCTURES = True```, or else ```bh1.mass``` and ```bh2.mass```.


Examples/BinaryBH/GNUmakefile
mirenradia, line 1: Can you remove the changes to this file?

I don't think they should be necessary.

mirenradia (dismissed):
Thanks for the PR, @ShaunSwain12.

I'll let others review this properly but I had one specific file comment below.

An alternative solution to this issue, which would remove the need for more conditional compilation in `BinaryBHLevel`, (which I know that Katy was not a huge fan of) is to get `TwoPuncturesInitialData` to instead write to the `bh1.mass` and `bh2.mass` parameters in the ParmParse table when its parameters are read.

Examples/BinaryBH/GNUmakefile
ShaunSwain12: Will remove these :)

ShaunSwain12: I've pushed the new changes, let me know if you have any other comments/suggestions.

Source/BlackHoles/TwoPuncturesInitialData.hpp
```diff
@@ -73,6 +73,25 @@ class TwoPuncturesInitialData
             s_two_punctures.center_offset[offset_dir] + s_two_punctures.par_b;
     }
 
+    void set_bh_masses()
```
KAClough: This name could be a bit confusing, maybe something like `add_bh_mass_params()` is better?

Source/BlackHoles/TwoPuncturesInitialData.hpp
```diff
@@ -73,6 +73,25 @@ class TwoPuncturesInitialData
...
+        // these masses need to be set for each restart too.
+
+        amrex::Real bh1_mass = s_two_punctures.give_bare_mass
+                                   ? s_two_punctures.par_m_minus
+                                   : s_two_punctures.target_M_minus;
+        amrex::Real bh2_mass = s_two_punctures.give_bare_mass
+                                   ? s_two_punctures.par_m_plus
+                                   : s_two_punctures.target_M_plus;
+
+        GRParmParse pp;
```
KAClough: I think strictly for consistency with other code you should do 

```
GRParmParse bh1_pp("bh1");
bh1_pp.queryAdd("mass", bh1_mass);

```
etc

Source/BlackHoles/TwoPuncturesInitialData.hpp
ShaunSwain12: Thanks Katy, I've made this change [here](https://github.com/GRTLCollaboration/GRTeclyn/commit/d46d851cc30bd8f0db7e084eb07c66350a61cedc).

Source/BlackHoles/TwoPuncturesInitialData.hpp
ShaunSwain12: Likewise, this change has been made [here](https://github.com/GRTLCollaboration/GRTeclyn/commit/d46d851cc30bd8f0db7e084eb07c66350a61cedc).

KAClough (approved):
LGTM

